PR #5, opened 2017-10-24
Move demo; clean up warnings

jkinkead:
Two commits here.

The first simply configures our compiler to stop complaining about the Sun interfaces used in the `tls` package.

The second moves the demo to a subproject, in preparation for moving it somewhere else. That step requires this library be published somewhere, though, so I'm not taking it yet. :)


andyday:
👍

PR #7, opened 2017-10-30
Cleanup various things

jkinkead:
Mostly checkstyle changes.

This looks like what's been done in the template out for review.

### DIFF
--- a/src/main/java/com/nordstrom/xrpc/http/Router.java
+++ b/src/main/java/com/nordstrom/xrpc/http/Router.java
@@ -16,17 +16,17 @@
 
 package com.nordstrom.xrpc.http;
 
-import com.nordstrom.xrpc.XConfig;
-import com.nordstrom.xrpc.tls.TLS;
-import com.nordstrom.xrpc.logging.MessageLogger;
-import com.nordstrom.xrpc.logging.ExceptionLogger;
-
 import static com.codahale.metrics.MetricRegistry.name;
-import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_LENGTH;
-import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_TYPE;
 import static io.netty.channel.ChannelOption.SO_BACKLOG;
 import static io.netty.channel.ChannelOption.SO_KEEPALIVE;
 import static io.netty.channel.ChannelOption.TCP_NODELAY;
+import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_LENGTH;
+import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_TYPE;
+
+import com.nordstrom.xrpc.XConfig;
+import com.nordstrom.xrpc.logging.ExceptionLogger;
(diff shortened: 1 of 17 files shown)


src/main/java/com/nordstrom/xrpc/http/Router.java
```diff
@@ -100,7 +101,7 @@
   // see http://metrics.dropwizard.io/3.2.2/getting-started.html for more on this
   private final MetricRegistry metrics = new MetricRegistry();
   private final Meter requests = metrics.meter("requests");
-  private final Histogram responseSizes = metrics.histogram(name(URLRouter.class, "responses"));
+  private final Histogram responseSizes = metrics.histogram(name(UrlRouter.class, "responses"));
```
emmanuel: I personally drank [Apple's Objective-C-](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingBasics.html#//apple_ref/doc/uid/20001281-1002931-BBCFHEAB) ([also Google's Obj-C guidelines](http://google.github.io/styleguide/objcguide.html#naming)) and [golang-](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) flavored Kool-aid, and find mixed case initialisms/acronyms to be distasteful. That said, I understand that they are the norm in Java, and therefore this will be the last that I remark on them in that context.

emmanuel:
The commit message says:

>Add Google's checkstyle template, minus one or two things.

For the record, what was omitted from Google's checkstyle template? And why? 

I'm not trying to cross-examine/interrogate, just want a quick note about how we're differing from upstream and why so the changes don't get lost later on when someone else updates to a newer release of the checkstyle template.

src/main/java/com/nordstrom/xrpc/http/Router.java
jkinkead: It's the norm, and enforced by the style guide. It was in the first Java style guide I read, published by Sun - it's been the official standard for a long time. [Kind](https://docs.oracle.com/javase/9/docs/api/java/net/URL.html) [of](https://docs.oracle.com/javase/9/docs/api/java/util/zip/GZIPOutputStream.html) [ironic](https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLSocket.html), [really](https://docs.oracle.com/javase/9/docs/api/java/security/cert/CRLSelector.html) - although [newer classes follow the standards](https://docs.oracle.com/javase/9/docs/api/java/security/DrbgParameters.html).

emmanuel: lol. That looks like there wasn't perfect agreement about the right way to name things. Consistency is hard, and naming things is specifically one of Martin Fowler's [two hard problems](https://www.martinfowler.com/bliki/TwoHardThings.html). Thanks for the links, it definitely helps me let myself off the hook to see that even Sun struggled with such things.

jkinkead:
I updated to use the newest version of the checkstyle template (matching the internal one). This includes documentation for what's been updated, as well as a revert-to-norm on the import ordering.

emmanuel:
LGTM.

PR #12, opened 2017-11-01
Add release plugin.

jkinkead:
For publishing to maven central.


jkinkead:
I'm adding instructions in a follow-up.

PR #14, opened 2017-11-02
Fix the latency histogram

xjdr:
Use the correct class (Timer) to measure request latency through the
pipeline and display the proper quantiles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 # Gradle output.
 /.gradle/
 build/
+*.jar
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 # Gradle output.
 /.gradle/
 build/
-*.jar
+/app.jar
--- a/bin/startTestServer.sh
+++ b/bin/startTestServer.sh
@@ -2,8 +2,9 @@
 
 PORT=8080
 
-mvn package && \
-  mv build/libs/*-all.jar app.jar
+
+./gradlew clean shadowJar && \
+  mv demo/build/libs/xrpc-demo-0.1.0-SNAPSHOT-all.jar app.jar && \
 
 java -ea          \
   $JAVA_OPTS                      \
--- a/bin/startTestServer.sh
+++ b/bin/startTestServer.sh
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 
+set -e
+
 PORT=8080
 
 
 ./gradlew clean shadowJar && \
   mv demo/build/libs/xrpc-demo-0.1.0-SNAPSHOT-all.jar app.jar && \
-
 java -ea          \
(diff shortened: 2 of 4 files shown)


src/main/java/com/nordstrom/xrpc/http/Router.java
```diff
@@ -96,9 +97,9 @@
   private Class<? extends ServerChannel> channelClass;
 
   // see http://metrics.dropwizard.io/3.2.2/getting-started.html for more on this
-  private final MetricRegistry metrics = new MetricRegistry();
+  private static final MetricRegistry metrics = new MetricRegistry();
   private final Meter requests = metrics.meter("requests");
-  private final Histogram responseSizes = metrics.histogram(name(UrlRouter.class, "responses"));
+  //  private final Histogram responseSizes = metrics.histogram(name(UrlRouter.class, "responses"));
```
jkinkead: Remove?

build.gradle
```diff
@@ -89,54 +90,68 @@ signing {
...
+//             name 'Jeff Rose'
+//             email 'jeff.rose@nordstrom.com'
+//           }
+//         }
+//       }
+//     }
+//   }
+// }
+
+publishing {
```
jkinkead: This doesn't look sufficient to publish to maven central.

build.gradle
```diff
@@ -24,6 +24,7 @@ release {
 apply plugin: 'checkstyle'
 apply plugin: 'java'
 apply plugin: 'maven'
+apply plugin: 'maven-publish'
```
jkinkead: Would you mind reverting this, or updating it to work with the release plugin + maven central?

[The Sonatype docs](http://central.sonatype.org/pages/gradle.html) only describe the plugin I was using, and I don't think the `maven-publish` will work without some extra configuration (signed jars, staging+regular repositories configured, license in POM).

bin/startTestServer.sh
```diff
@@ -2,8 +2,9 @@
 
 PORT=8080
 
-mvn package && \
-  mv build/libs/*-all.jar app.jar
+
+./gradlew clean shadowJar && \
+  mv demo/build/libs/xrpc-demo-0.1.0-SNAPSHOT-all.jar app.jar && \
```
jkinkead: This looks broken - unless the escape carries past the blank line below?

I would also consider `set -e`, which @emmanuel pointed me to yesterday. :)

.gitignore
```diff
@@ -1,3 +1,4 @@
 # Gradle output.
 /.gradle/
 build/
+*.jar
```
jkinkead: I'm not a big fan of build scripts that cruft up non-build directories.

Also, this needs to be `/*.jar` or better `/app.jar`, so that it doesn't ignore the bundled `gradlew` jar.

PR #21, opened 2017-11-07
Add Http/2 Server

xjdr:
This is a basic implementation of Http/2 which will auto-negotiate and
should return the same value for Http/1.1 and Http/2 requests. This commit
does not handle anything in the request body as of yet but is a good
starting point.


src/main/java/com/nordstrom/xrpc/http/Context.java
```diff
@@ -16,20 +16,37 @@
...
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.handler.codec.http.FullHttpRequest;
+import io.netty.handler.codec.http2.Http2DataFrame;
+import io.netty.handler.codec.http2.Http2Headers;
 import lombok.Getter;
+import lombok.Setter;
 
 /** Request context. */
 public class Context {
```
pdex: # Proposal: Rename this class
I don't want to bikeshed about class names but as we appear to be making our own Request class here. That's fine, name it as such. Netty's request/response objects have a lot of sharp edges that you will cut yourself on, we should be trying to smooth over those rough edges for the user.

# Proposal: Provide a single request object to the user.
I'm fine with stuffing both the http1 and http2 request objects into this class for now, but we need to put some time into thinking about how to abstract over those use cases into a single Request/Response model. The user probably doesn't care about the protocol 99% of the time, they just want to get a request and produce a response. Users who are used to working with http1 are going to be confused by the fact that with http2 they now get a function call with a headers object followed by one or more function calls with a data object.

# Footnote: Large payloads
This Function<Request, Response> model that we're building right now will work fine for reasonable payloads, but the http1 [HttpObjectAggregator](https://netty.io/4.1/api/io/netty/handler/codec/http/HttpObjectAggregator.html)
will send a 4xx class `HttpResponse` if the payload is too large. The http2 model doesn't have this issue since it expects the user to consume a headers object followed by one or more data objects. When we're thinking about how to abstract over these two protocols we should keep large payloads in mind.

src/main/java/com/nordstrom/xrpc/http/Router.java
```diff
@@ -238,28 +272,51 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
         response.headers().set(CONTENT_TYPE, "text/plain");
         response.headers().setInt(CONTENT_LENGTH, 0);
         ctx.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE);
       }
+
+      if (msg instanceof Http2Request) {
+        log.info("Is a Http2 Request");
+        Http2Request<Http2Headers> request = (Http2Request) msg;
+        String uri = request.payload.path().toString();
+        for (Route route : routes.keySet()) {
```
pdex: @xjdr this is why you had an ordered map.

src/main/java/com/nordstrom/xrpc/http/Router.java
```diff
@@ -84,8 +83,9 @@
 
   private final int bossThreadCount;
   private final int workerThreadCount;
-  /** LinkedHashMap to retain ordering. */
-  private final Map<Route, Handler> routes = new LinkedHashMap<>();
+  // TODO(JR): Not sure why we would want to retain ordering here, but added a high
```
pdex: @xjdr scroll down.

src/main/java/com/nordstrom/xrpc/http/Context.java
xjdr: One potential solution would be to remove the Http1 Aggregator and abstract over the http multipart message in a similar fashion to the http2 data frame. While more complicated, it may provide a more consistent implementation pattern.

src/main/java/com/nordstrom/xrpc/tls/Tls.java
```diff
@@ -72,11 +54,20 @@ public Tls(String cert, String key) {
     this.sslCtx = buildEncryptionHandler();
   }
 
-  public ChannelHandler getEncryptionHandler() {
+  public static X509Certificate createSelfSigned() {
```
pdex: [netty has a cert generator](https://netty.io/4.0/api/io/netty/handler/ssl/util/SelfSignedCertificate.html) Any objection to using that?

.gitignore
```diff
@@ -2,3 +2,7 @@
 /.gradle/
 build/
 /app.jar
+.idea/
```
jkinkead: This should go in a global .gitignore.

demo/src/main/java/com/nordstrom/xrpc/demo/Example.java
```diff
@@ -76,8 +76,10 @@ public static void main(String[] args) {
     // Define a simple function call.
     Handler peopleHandler =
         context -> {
-          return Recipes.newResponseOk(
-              adapter.toJson(people), Recipes.ContentType.Application_Json);
+          return Recipes.newResponse(
+            HttpResponseStatus.OK,
+            context.getAlloc().directBuffer().writeBytes(adapter.toJson(people).getBytes()),
```
jkinkead: If there's only one or two methods we ever want users calling on `alloc`, should we just expose those directly in the `Context` class, as like `getDirectBuffer` or whatever?

src/main/java/com/nordstrom/xrpc/http/Context.java
pdex: I would argue that we should target small payloads that can be represented as Request/Response objects. Forcing the user to think about this function call has a header and this function call has a body will get very ugly. Furthermore the Router constructor should take a max payload size as an argument: 
```java
    // Build your router
    Router router = new Router(maxPayloadSize);
```

Large payloads that require multiple function calls should be the exceptional case. This allows us to focus on something easy to understand now and we can deal with the outlier cases when they come up.

src/main/java/com/nordstrom/xrpc/http/Http2FrameForwarder.java
```diff
@@ -0,0 +1,108 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http2.DefaultHttp2DataFrame;
+import io.netty.handler.codec.http2.Http2Exception;
+import io.netty.handler.codec.http2.Http2Flags;
+import io.netty.handler.codec.http2.Http2FrameListener;
+import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.handler.codec.http2.Http2Settings;
+
+/**
+ * Forwards the frame with stream id to the next handler in the pipeline
```
jkinkead: Linter says: Javadoc comment should have terminal period.

src/main/java/com/nordstrom/xrpc/http/Http2FrameForwarder.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+public class Http2FrameForwarder implements Http2FrameListener {
+
+  private final boolean isServer;
+
+  public Http2FrameForwarder(boolean isServer) {
+    this.isServer = isServer;
+  }
+
+  public static Http2FrameForwarder create(boolean isServer) {
```
jkinkead: This just forwards to the constructor; remove or make constructor private?

src/main/java/com/nordstrom/xrpc/http/Http2FrameForwarder.java
```diff
@@ -0,0 +1,108 @@
...
+import io.netty.handler.codec.http2.Http2Settings;
+
+/**
+ * Forwards the frame with stream id to the next handler in the pipeline
+ */
+public class Http2FrameForwarder implements Http2FrameListener {
+
+  private final boolean isServer;
+
+  public Http2FrameForwarder(boolean isServer) {
```
jkinkead: Document parameter / member? It's not obvious what this means.

src/main/java/com/nordstrom/xrpc/http/Http2FrameForwarder.java
```diff
@@ -0,0 +1,108 @@
...
+    // TODO(CK): We don't currently have a use case for these frames
+  }
+
+  @Override
+  public void onSettingsRead(ChannelHandlerContext ctx, Http2Settings settings) throws Http2Exception {
+    // TODO(CK): We don't currently have a use case for these frames
+  }
+
+  @Override
+  public void onPingRead(ChannelHandlerContext ctx, ByteBuf data) throws Http2Exception {
```
jkinkead: Does netty handle the ACK here? I would worry if we're not ACK-ing properly.

src/main/java/com/nordstrom/xrpc/http/Http2FrameForwarder.java
```diff
@@ -0,0 +1,108 @@
...
+    // TODO(CK): We don't currently have a use case for these frames
+  }
+
+  @Override
+  public void onPushPromiseRead(ChannelHandlerContext ctx, int streamId, int promisedStreamId, Http2Headers headers, int padding) throws Http2Exception {
+    // TODO(CK): We don't currently have a use case for these frames
+  }
+
+  @Override
+  public void onGoAwayRead(ChannelHandlerContext ctx, int lastStreamId, long errorCode, ByteBuf debugData) throws Http2Exception {
```
jkinkead: Does netty handle closing the connection here? We should try to shut down quickly & gracefully here.

src/main/java/com/nordstrom/xrpc/http/Http2Handler.java
```diff
@@ -0,0 +1,88 @@
...
+    ctx.channel().attr(STREAM_ID_KEY).set(streamId);
+  }
+
+  private int currentStreamId;
+
+  public Http2Handler(Http2ConnectionDecoder decoder, Http2ConnectionEncoder encoder, Http2Settings initialSettings) {
+    super(decoder, encoder, initialSettings);
+  }
+
+  private void writeHeaders(ChannelHandlerContext ctx, Http2Headers headers, boolean eos, ChannelPromise promise) throws Exception {
```
jkinkead: Linter says: Don't throw `Exception` (here and elsewhere).

src/main/java/com/nordstrom/xrpc/http/Http2HandlerBuilder.java
```diff
@@ -0,0 +1,42 @@
+package com.nordstrom.xrpc.http;
+
+import static io.netty.handler.logging.LogLevel.INFO;
+
+import io.netty.handler.codec.http2.*;
```
jkinkead: Linter says: No wildcard imports.

src/main/java/com/nordstrom/xrpc/http/Http2HandlerBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.function.Function;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class Http2HandlerBuilder extends AbstractHttp2ConnectionHandlerBuilder<Http2Handler, Http2HandlerBuilder> {
```
jkinkead: It wasn't feeling like Java until we started extending some abstract factories.

src/main/java/com/nordstrom/xrpc/http/Context.java
jkinkead: We should have the framework support large payloads somehow. If we don't, we'll be forcing users to cobble together their own solution, or migrate to a new framework.

On a recent project, we abandoned an otherwise-decent HTTP server framework largely because it didn't support multipart messages, and downloading large-ish data files was a primary use-case.

andyday: other frameworks use a stream input or output to signal a large chunked request/response maybe we could do something similar but not sure what exactly.  also 👍 to providing our own request/response with all rough edges smoothed.  this should include friendly access to all the parts of the request (uri params, query params, etc) and might even provide friendly methods for marshalling/unmarshalling...

demo/src/main/java/com/nordstrom/xrpc/demo/Example.java
```diff
@@ -94,7 +96,7 @@ public static void main(String[] args) {
         context -> {
           // TODO(jkinkead): Clean this up; we should have a helper to handle this.
           Dino output = dinos.get(0);
-          ByteBuf bb = Unpooled.compositeBuffer();
+          ByteBuf bb = context.getAlloc().compositeDirectBuffer();
```
andyday: yuck.  do we want users messing directly with ByteBuf?

demo/src/main/java/com/nordstrom/xrpc/demo/Example.java
andyday: I'd prefer to do something like

```java
return Response.Ok(...)
// or
return Response.BadRequest(...)
```

over

```java
Recipes.newResponse(...)
```

src/main/java/com/nordstrom/xrpc/http/Http2Handler.java
```diff
@@ -0,0 +1,88 @@
...
+import io.netty.handler.codec.http2.Http2Settings;
+import io.netty.util.AttributeKey;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class Http2Handler extends Http2ConnectionHandler {
+
+  private static final AttributeKey<Integer> STREAM_ID_KEY = AttributeKey.newInstance("xrpc_h2_stream_id");
+
+  public static int defaultValue(Integer i) {
```
andyday: this is a common utility should be part of a utility class somewhere...

src/main/java/com/nordstrom/xrpc/http/Router.java
```diff
@@ -228,6 +225,43 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
+      this.router = router;
+    }
+
+    @Override
+    protected void configurePipeline(ChannelHandlerContext ctx, String protocol) throws Exception {
+      if (ApplicationProtocolNames.HTTP_2.equals(protocol)) {
+        log.info("Using Http/2");
+        ChannelPipeline cp = ctx.pipeline();
+        cp.addLast(new Http2HandlerBuilder().server(true).build());
+        cp.addLast("routingFiler", router);
```
andyday: sp.  should be routingFilter

src/main/java/com/nordstrom/xrpc/http/Router.java
andyday: not sure a map even makes sense at all.  should be a key/value list

src/main/java/com/nordstrom/xrpc/tls/Tls.java
```diff
@@ -134,30 +125,52 @@ public SslContext buildEncryptionHandler() {
...
-        | InvalidKeyException e) {
+      | KeyStoreException
+      | UnrecoverableKeyException
+      | CertificateException
+      | NoSuchProviderException
+      | IllegalArgumentException
+      | IOException
+      | SignatureException
+      | InvalidKeyException e) {
       e.printStackTrace();
```
andyday: what do we do with all the other exceptions?

andyday:
given the significance of this change, we should probably have unit or functional tests for it...

src/main/java/com/nordstrom/xrpc/tls/Tls.java
xjdr: While not a very satisfactory answer, there was a very specific reason I chose to implement my own several years ago starting with xjdr/xio. I cannot for the life of me remember what it is, but i will go looking through the old xjdr/xio commit logs to see if i can dig up a better explanation.

demo/src/main/java/com/nordstrom/xrpc/demo/Example.java
xjdr: I don't disagree on the naming, but probably should be encapsulated in a different commit.

src/main/java/com/nordstrom/xrpc/http/Handler.java
```diff
@@ -18,10 +18,8 @@
 
 import io.netty.handler.codec.http.HttpResponse;
 
-import java.io.IOException;
-
 /** A handler for an HTTP route. */
 @FunctionalInterface
 public interface Handler {
-  HttpResponse handle(Context context) throws IOException;
+  HttpResponse handle(XrpcRequest xrpcRequest);
```
jkinkead: Any reason for removing the exception? This let lambdas throw an exception; otherwise, you have to catch everything locally.

I don't think we want to force users to handle exceptions in the handler. It's much easier to have a higher-level logger & 50X response server.

src/main/java/com/nordstrom/xrpc/http/Router.java
```diff
@@ -78,51 +56,57 @@
...
-  private final Map<Route, Handler> routes = new LinkedHashMap<>();
-
-  private Channel channel;
-  private EventLoopGroup bossGroup;
-  private EventLoopGroup workerGroup;
-  private Class<? extends ServerChannel> channelClass;
-
-  // see http://metrics.dropwizard.io/3.2.2/getting-started.html for more on this
-  private static final MetricRegistry metrics = new MetricRegistry();
+  // TODO(JR): Not sure why we would want to retain ordering here, but added a high
```
jkinkead: Users assume ordering. Without ordering, the following is nondeterministic:
```java
router.addRoute("/people/admin", adminHandler);
router.addRoute("/people/{person}", personHandler);
```

src/main/java/com/nordstrom/xrpc/http/Router.java
```diff
@@ -78,51 +56,57 @@
...
-
   private final Tls tls;
+  private Channel channel;
+  private EventLoopGroup bossGroup;
+  private EventLoopGroup workerGroup;
+  private Class<? extends ServerChannel> channelClass;
 
-  private static ThreadFactory threadFactory(String nameFormat) {
-    return new ThreadFactoryBuilder().setNameFormat(nameFormat).build();
+  public Router(XConfig config) {
```
jkinkead: Reimplement as an invocation of the other constructor?

src/main/java/com/nordstrom/xrpc/http/XrpcRequest.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        return h1Request;
+      } catch (Http2Exception e) {
+        //TODO(JR): Do something more meaningful with this exception
+        e.printStackTrace();
+      }
+    }
+
+    return null;
```
jkinkead: I'd throw an `IllegalStateException` here - or else at the beginning of the function, if both requests are `null`.

src/main/java/com/nordstrom/xrpc/http/XrpcRequest.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Returns the variable with the given name, or null if that variable doesn't exist.
+   */
+  public String variable(String name) {
+    return groups.get(name);
+  }
+
+  /**
+   * Create a convenience function to prevent direct access too the Allocator
```
jkinkead: s/too/to/

.gitignore
xjdr: I agree and am happy to pull this out but for an OSS release I would advocate for leaving this in JIC.

andyday: agreed.  i tend to favor project based ignores over any global setting...  it keeps everyone on the same page.

src/main/java/com/nordstrom/xrpc/http/XrpcRequest.java
xjdr: Good Callout. I will update to reflect that.

src/main/java/com/nordstrom/xrpc/http/Router.java
andyday: not all frameworks assume ordering.  especially those that use annotations to declare routes.   usually route is determined by ordering on specificity.   `/people/admin` is more specific so if it matches it is chosen over `/people/{person}`.

.gitignore
jkinkead: Mostly I mean that you shouldn't need to ever submit this, because it should be in your global `.gitignore`, alongside OS-specific / tool-specific cruft. Once we start adding stuff like this, [the](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) [list](https://github.com/github/gitignore/blob/master/Global/Emacs.gitignore) [grows](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) [huge](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore).

src/main/java/com/nordstrom/xrpc/http/Router.java
jkinkead: Sure, but it's nondeterministic here. I wasn't asserting that it HAD to be in-declaration-order, just consistent.

.gitignore
andyday: understood.  i would argue that .idea has become commonplace enough to make it in here.  as an example, netty project uses intellij and has .idea in its .gitignore as do other popular oss projects.

src/main/java/com/nordstrom/xrpc/http/Router.java
andyday: gotcha...   good callout...

.gitignore
jkinkead: My point is more that: If you're adding ignores for tool-specific files, you probably have git misconfigured on your end. I don't object to this being here, but I object to what it implies. :P

src/main/java/com/nordstrom/xrpc/http/XrpcRequest.java
xjdr: sticky keys ...

src/main/java/com/nordstrom/xrpc/http/Handler.java
xjdr: Throwing here would potentially have the exception caught by the Netty pipeline vs. having the user deal with it in a more programmatic fashion in their app code. I agree that abstracting the exception handling is desireable, i think this could cause a greater degree of confusion. For example:

```java
    private void writeResponse(ChannelHandlerContext ctx, int streamId, Route route) {
      FullHttpResponse h1Resp = (FullHttpResponse) routes.get(route).handle(xrpcRequest);
      Http2Headers responseHeaders = HttpConversionUtil.toHttp2Headers(h1Resp, true);
      Http2DataFrame responseDataFrame = new DefaultHttp2DataFrame(h1Resp.content(), true);
      encoder().writeHeaders(ctx, streamId, responseHeaders, 0, false, ctx.newPromise());
      encoder().writeData(ctx, streamId, responseDataFrame.content(), 0, true, ctx.newPromise());
    }
``` 
The above logic is in the pipeline code and the Overridden  method does not throw. If we had the function re-throw, the error would most likely get swallowed by netty.

jkinkead: Wouldn't it make sense for us to have a `try/catch` in the `writeResponse` method, around the call to `handle`?

I find checked exceptions to impede functional programming. Without a way to bubble out exceptions, the user is forced to either handle each exception separately within a lambda, resulting in ugly repeated code [like this](https://github.com/Nordstrom/xrpc/blob/7fa0f0b90b0ebd409ecdbd5b3f8e230805070108/demo/src/main/java/com/nordstrom/xrpc/demo/Example.java#L119-L137), or else write a separate handler wrapper that does the `try/catch` for them.

.gitignore
andyday: for oss projects the audience and potential contributions is much wider.  safe guards are a good thing in this case...

demo/src/main/java/com/nordstrom/xrpc/demo/Example.java
manimaul: @andyday - i'd rather stick to standard java method naming: https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

andyday: @manimaul agreed this should be more like

```java
return Response.ok(...)
// or
return Response.badRequest(...)
```

i should have remembered that, but i jump from language to language these days so this one slid through...

PR #22, opened 2017-11-08
Adds generatedFilesBaseDir to demo build.gradle.

spenserpothier:
* Import statement in the demo was not resolving during build. This is
  to fix that problem.

### DIFF
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -50,6 +50,7 @@ protobuf {
   protoc {
     artifact = 'com.google.protobuf:protoc:3.0.0'
   }
+  generatedFilesBaseDir = "$projectDir/src/"
 }
 
 shadowJar {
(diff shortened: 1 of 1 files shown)


demo/build.gradle
```diff
@@ -50,6 +50,7 @@ protobuf {
   protoc {
     artifact = 'com.google.protobuf:protoc:3.0.0'
   }
+  generatedFilesBaseDir = "$projectDir/src/"
```
jkinkead: This isn't correct - the generated files should be put in with the build artifacts, not the source. The problem here is with IntelliJ not fully-handling the gradle build (this works find on the commandline). [Per this issue](https://github.com/google/protobuf-gradle-plugin/issues/109), IntelliJ is only looking at the sources sets gradle defines.

You should apply the solution at the top of that bug report.

PR #23, opened 2017-11-21
0.1.1

spenserpothier:


src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+            buildRetryLoop());
+
+    Futures.addCallback(
+        connectFuture,
+        new FutureCallback<ChannelFuture>() {
+          @Override
+          public void onSuccess(ChannelFuture result) {
+            try {
+              Channel channel = result.await().channel();
+              channel.writeAndFlush(request);
```
jkinkead: It makes me very nervous to be referencing a non-final field from a future.

src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+  private FullHttpRequest request = null;
+
+  public Call(XrpcClient client, String uri) {
+
+    this.client = client;
+    this.uri = uri;
+  }
+
+  public Call get(FullHttpRequest request) {
+    this.request = request;
```
jkinkead: Hrm. The way that this API is set up, it looks like one could create a `Call` instance to share among multiple threads - but a mutable call like this prevents that from happening, and makes this class not thread-safe.

Either this should return a new `Call` instance with `request` set (and `request` should be final), or else `execute` should take the request as input, and `get` should be removed.

(also, one tiny style nit: calling this `get` in an HTTP context makes me think this is a GET request; but it's not)

src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public class Call {
+  private final XrpcClient client;
+  private final String uri;
+
+  private FullHttpRequest request = null;
+
+  public Call(XrpcClient client, String uri) {
```
jkinkead: It looks like the intent is to always create this via the `newCall` method in `XrpcClient`. If that's the case, this should probably be `protected` or default-access (package-private).

src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+
+    this.client = client;
```
jkinkead: This is only used for its `Bootstrap` member. If you make the constructor non-public, consider making this just a bootstrap instance.

src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+import io.netty.channel.ChannelFutureListener;
+import io.netty.handler.codec.http.FullHttpRequest;
 import io.netty.handler.codec.http.FullHttpResponse;
+import java.net.InetSocketAddress;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public class Call {
```
jkinkead: It would be nice to have Javadoc on the public API points.

src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+
+    if (response.isCancelled()) {
+      return error;
+    }
+
+    return response;
+  }
+
+  private RetryLoop buildRetryLoop() {
+    return buildRetryLoop(50, 500, 4);
```
jkinkead: Make these configurable somehow? Or maybe a TODO to do the same?

src/main/java/com/nordstrom/xrpc/client/retry/TracerDriver.java
```diff
@@ -0,0 +1,39 @@
...
+/** Mechanism for timing methods and recording counters */
+public interface TracerDriver {
+  /**
+   * Record the given trace event
+   *
+   * @param name of the event
+   * @param time time event took
+   * @param unit time unit
+   */
+  public void addTrace(String name, long time, TimeUnit unit);
```
jkinkead: Use a default no-op implementation, for easy overriding? You should also remove the redundant `public` modifiers on this.

src/main/java/com/nordstrom/xrpc/client/retry/RetryLoop.java
```diff
@@ -0,0 +1,88 @@
...
+  public RetryLoop(RetryPolicy retryPolicy, AtomicReference<TracerDriver> tracer) {
+    this.retryPolicy = retryPolicy;
+    this.tracer = tracer;
+  }
+
+  public static RetrySleeper getDefaultRetrySleeper() {
+    return sleeper;
+  }
+
+  public static boolean shouldRetry(int rc) {
```
jkinkead: This looks unused. Remove?

src/main/java/com/nordstrom/xrpc/client/retry/RetryLoop.java
```diff
@@ -0,0 +1,88 @@
...
+      new RetrySleeper() {
+        @Override
+        public void sleepFor(long time, TimeUnit unit) throws InterruptedException {
+          unit.sleep(time);
+        }
+      };
+
+  private final long startTimeMs = System.currentTimeMillis();
+  private final RetryPolicy retryPolicy;
+  private final AtomicReference<TracerDriver> tracer;
```
jkinkead: It's confusing that you're using an atomic reference given that the rest of this class isn't thread-safe.

src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+  }
+
+  private RetryLoop buildRetryLoop(int baseSleep, int maxSleep, int reties) {
+    BoundedExponentialBackoffRetry retry =
+        new BoundedExponentialBackoffRetry(baseSleep, maxSleep, reties);
+
+    TracerDriver tracerDriver =
```
jkinkead: Seems like this could use a TODO to flesh out?

src/main/java/com/nordstrom/xrpc/client/HttpResponseHandler.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.handler.codec.http.FullHttpResponse;
+
+public class HttpResponseHandler extends SimpleChannelInboundHandler<FullHttpResponse> {
+  SettableFuture<FullHttpResponse> responseFuture = SettableFuture.create();
+
+  public HttpResponseHandler() {}
```
jkinkead: Remove unneeded empty constructor?

src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -0,0 +1,126 @@
...
+    if (!matcher.find()) {
+      url = "http://" + url;
+    }
+    return url;
+  }
+
+  public static List<String> getDomainLevels(String host) {
+    Preconditions.checkNotNull(host);
+
+    // Automatically include www prefix if not present.
```
jkinkead: This . . . confuses me.

src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -0,0 +1,126 @@
...
+  public static String addProtocol(String url) {
+    Preconditions.checkNotNull(url);
+    Matcher matcher = URL_PROTOCOL_REGEX.matcher(url);
+    if (!matcher.find()) {
+      url = "http://" + url;
+    }
+    return url;
+  }
+
+  public static List<String> getDomainLevels(String host) {
```
jkinkead: Unused?

src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -0,0 +1,126 @@
...
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class XUrl {
```
jkinkead: What about [HttpUrl](https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.html)? Or at least wrapping it instead of wrapping the (fairly terrible) java.net versions?

src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -1,12 +1,159 @@
...
+  }
 
   public ListenableFuture<FullHttpResponse> execute() {
+    Preconditions.checkState(request != null);
+    final SettableFuture<FullHttpResponse> error = SettableFuture.create();
+    final SettableFuture<FullHttpResponse> response = SettableFuture.create();
+
+    ListenableFuture<ChannelFuture> connectFuture =
+        connect(
+            new InetSocketAddress(XUrl.getHost(uri), XUrl.getPort(uri)),
```
jkinkead: This parses the URL twice, FWIW.

src/main/java/com/nordstrom/xrpc/client/retry/SleepingRetry.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.concurrent.TimeUnit;
+
+abstract class SleepingRetry implements RetryPolicy {
+  private final int n;
+
+  protected SleepingRetry(int n) {
+    this.n = n;
+  }
+
+  // made public for testing
```
jkinkead: Default access should be sufficient for this.

Consider using Guava's `@VisibleForTesting` annotation instead of a comment.

src/main/java/com/nordstrom/xrpc/client/retry/SleepingRetry.java
```diff
@@ -0,0 +1,47 @@
...
+  public int getN() {
+    return n;
+  }
+
+  public boolean allowRetry(int retryCount, long elapsedTimeMs, RetrySleeper sleeper) {
+    if (retryCount < n) {
+      try {
+        sleeper.sleepFor(getSleepTimeMs(retryCount, elapsedTimeMs), TimeUnit.MILLISECONDS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
```
jkinkead: Maybe comment that you're force-setting the interrupt status, if that's what you're doing?

Usually I see something more like:

```java
} catch (InterruptedException e) {
  if (Thread.currentThread().isInturrupted()) {
    // Non-spurious interrupt.
    return false;
  } else {
    // Ignore.
  }
}
```

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public final class Http2Handler extends Http2ConnectionHandler implements Http2FrameListener {
+
+  private final Meter requests;
+  private final AtomicReference<ImmutableSortedMap<Route, Handler>> routes;
+  private XrpcRequest xrpcRequest;
```
jkinkead: Is a new instance of this class created per request, then?

src/main/java/com/nordstrom/xrpc/client/retry/RetryLoop.java
xjdr: The Tracer may be used in multiple places once the circuit breaker / retry grey list is in place

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
xjdr: Yes, that the netty default if `@ChannelHander.Sharable` is not in the class annotation. We try to keep object creation to a minimum here, but in this instance, it is easier to create a a new object per call.

PR #35, opened 2017-12-08
Ulr and pipeline fixes

xjdr:
This commit should address a number of open issues. The primary addition is the HttpMethod to  addRoute() which allows you to bind multiple handlers to the same path and sort them by method.

### DIFF
--- a/src/main/java/com/nordstrom/xrpc/client/XUrl.java
+++ b/src/main/java/com/nordstrom/xrpc/client/XUrl.java
@@ -17,12 +17,19 @@ package com.nordstrom.xrpc.client;
  */
 
 import com.google.common.base.Preconditions;
+import lombok.extern.slf4j.Slf4j;
+
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import lombok.extern.slf4j.Slf4j;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toList;
 
 @Slf4j
 public class XUrl {
@@ -99,7 +106,7 @@ public class XUrl {
   }
 
   private static final Pattern URL_PROTOCOL_REGEX =
-      Pattern.compile("^https?://", Pattern.CASE_INSENSITIVE);
+    Pattern.compile("^https?://", Pattern.CASE_INSENSITIVE);
 
   public static String addProtocol(String url) {
     Preconditions.checkNotNull(url);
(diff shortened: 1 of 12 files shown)


src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -17,13 +17,20 @@
  */
 
 import com.google.common.base.Preconditions;
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.*;
```
jkinkead: Why does IntelliJ do this? It's ALWAYS been considered poor coding style.

src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -109,4 +116,39 @@ public static String addProtocol(String url) {
...
+          return new InetSocketAddress(uri.getHost(), 80);
+        } else {
+          return new InetSocketAddress(uri.getHost(), 443);
+        }
+      } else {
+        return new InetSocketAddress(uri.getHost(), uri.getPort());
+      }
+    } catch (URISyntaxException e) {
+      log.info("Malformed url: " + url);
+      //TODO(JR): Should this rethrow
```
jkinkead: I vote yes.

I'd be happy with it throwing without logging, too - this is going to virtually always be unrecoverable.

src/main/java/com/nordstrom/xrpc/server/http/XHttpMethod.java
```diff
@@ -0,0 +1,162 @@
...
+   * This specification reserves the method name CONNECT for use with a proxy that can dynamically
+   * switch to being a tunnel
+   */
+  public static final XHttpMethod CONNECT = new XHttpMethod("CONNECT");
+
+  /**
+   * This specification reserves the method name ANY for use with a the UrlRoute to allow matching
+   * of any HttpMethod
+   */
+  public static final XHttpMethod ANY = new XHttpMethod("ANY");
```
jkinkead: What if you used a string that was an impossible HTTP method name, like `" ANY"` (leading space)?

src/test/java/com/nordstrom/xrpc/client/XUrlTest.java
```diff
@@ -0,0 +1,63 @@
+package com.nordstrom.xrpc.client;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class XUrlTest {
```
jkinkead: I'd add a test for URLs without params, with empty paths, and with a path terminating in a slash.

Stuff like:
```java
String noPath = "https://api.nordstrom.com";
String noPathTrailingSlash = "https://api.nordstrom.com/";
String noPathQuery = "https://api.nordstrom.com?foo=bar";
String withPathNoQuery = "https://api.nordstrom.com/v1";
String withPathTrailingSlashNoQuery = "https://api.nordstrom.com/v1/";
String withPathTrailingSlash = "https://api.nordstrom.com/v1/?foo=bar";
```

src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -109,4 +116,39 @@ public static String addProtocol(String url) {
     }
     return url;
   }
+
+  public static Map<String, List<String>> decodeQString(String url) {
```
jkinkead: For a public API, I'd expand Q -> Query.

src/main/java/com/nordstrom/xrpc/client/XUrl.java
xjdr: It happens for more than 8 imports of the same package. I will try to hunt down the setting and fix. I hate it as well ...

src/main/java/com/nordstrom/xrpc/server/http/XHttpMethod.java
xjdr: It will let you create it but it will never match as it isn't a valid verb. It's really just a glorified enum, but we could add error checking and input validation if this becomes a problem.

PR #46, opened 2017-12-20
Add support for KQueue

xjdr:
Add support for the KQueue eventing system for OSX development.

This also bumps the version of netty to 4.1.19


src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -156,7 +156,19 @@ public void operationComplete(ChannelFuture future) {
           }
         };
 
-    bootstrap.connect(server).addListener(listener);
+    //TODO(JR): There should be a timeout here
+    try {
+      bootstrap.connect(server).addListener(listener).await(200, TimeUnit.MILLISECONDS);
+    } catch (InterruptedException e) {
```
jkinkead: This means that ctrl-C won't actually kill this, right? That's pretty annoying behavior.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -236,18 +240,27 @@ public void listenAndServe() throws IOException {
...
-      b.option(SO_BACKLOG, 1024);
-      b.childOption(SO_KEEPALIVE, true);
-      b.childOption(TCP_NODELAY, true);
     }
 
+    b.option(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT);
+    b.option(ChannelOption.SO_BACKLOG, 8192);
+    b.option(ChannelOption.SO_REUSEADDR, true);
+
+    b.childOption(ChannelOption.SO_REUSEADDR, true);
```
jkinkead: (not knowing anything about the underlying code) Is this not inherited from the parent?

src/main/java/com/nordstrom/xrpc/client/Call.java
xjdr: ctrl-c should still work as it will kill the thread pool, etc

jkinkead: Usually that just triggers an `InterruptedException` . . . but yeah, that's on the main thread. So we should be good. :)

pdex: The JVMs relationship with POSIX signals is bizarre at best. I think CTRL-c should work unless you've gone out of your way to disable it.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -236,18 +240,27 @@ public void listenAndServe() throws IOException {
       bossGroup = new EpollEventLoopGroup(bossThreadCount, threadFactory(workerNameFormat));
       workerGroup = new EpollEventLoopGroup(workerThreadCount, threadFactory(workerNameFormat));
       channelClass = EpollServerSocketChannel.class;
+    } else if (KQueue.isAvailable()) {
+      log.info("Using KQueue");
+      bossGroup = new KQueueEventLoopGroup(bossThreadCount, threadFactory(workerNameFormat));
+      workerGroup = new KQueueEventLoopGroup(workerThreadCount, threadFactory(workerNameFormat));
+      channelClass = KQueueServerSocketChannel.class;
+      b.option(EpollChannelOption.SO_REUSEPORT, true);
     } else {
```
pdex: I kinda want to see all of this hidden away in another class. I'll make an issue.

src/main/java/com/nordstrom/xrpc/client/Call.java
jkinkead: Yeah, I think I'm confusing another language's behavior here - or maybe some I/O operations do this. I can't reproduce with simple `wait()` / `notify()` blocks.

andyday: you might be thinking of terraform ;)  jk...

Issue #47, opened 2017-12-20
Move the channel implementation switch out of the Router


pdex:
https://github.com/Nordstrom/xrpc/pull/46#pullrequestreview-84912915

#46 

Include a link to netty native docs in the new class:
https://netty.io/wiki/native-transports.html

xjdr:
457ba03bab3a2113e192edea4f381bd7538d9db6

Issue #43, opened 2017-12-15
refactor XrpcChannelContext into server state and channel handler classes


pdex:
https://github.com/Nordstrom/xrpc/pull/38/files#r157036367
[#38]

xjdr:
Addressed by 457ba03bab3a2113e192edea4f381bd7538d9db6

PR #49, opened 2017-12-22
Removed XrpcChannelContext to address Issue 43

xjdr:
Removed the XrpcChannelContext and added an attribute to
ctx.channel().attr() to address Issue 43 as well as remove a ~250us
regression in the code base. This commit should generate significantly less
garbage and has removed the added regression. To reproduce run

```bash
$ wrk -c 15 -d 60 -t 4 https://localhost:8080/people --latency
Running 1m test @ https://localhost:8080/people
  4 threads and 15 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   446.38us    1.26ms 136.34ms   99.03%
    Req/Sec   127.62     31.13     0.90k    99.08%
  Latency Distribution
     50%  390.00us
     75%  451.00us
     90%  548.00us
     99%    1.63ms
  30525 requests in 1.00m, 2.10MB read
```

### DIFF
--- a/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
+++ b/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
@@ -10,6 +10,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http.*;
 import io.netty.handler.codec.http2.*;
+import io.netty.util.AttributeKey;
 import java.io.IOException;
 import java.util.Map;
 import java.util.Optional;
@@ -17,17 +18,15 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public final class Http2Handler extends Http2ConnectionHandler implements Http2FrameListener {
-
-  private final XrpcChannelContext xctx;
-  private XrpcRequest xrpcRequest;
+  private static final AttributeKey<XrpcConnectionContext> CONNECTION_CONTEXT =
+      AttributeKey.valueOf("XrpcConnectionContext");
+  private static final AttributeKey<XrpcRequest> XRPC_REQUEST = AttributeKey.valueOf("XrpcRequest");
 
   Http2Handler(
-      XrpcChannelContext xctx,
       Http2ConnectionDecoder decoder,
       Http2ConnectionEncoder encoder,
       Http2Settings initialSettings) {
     super(decoder, encoder, initialSettings);
-    this.xctx = xctx;
   }
 
   @Override
@@ -42,12 +41,14 @@ public final class Http2Handler extends Http2ConnectionHandler implements Http2F
 
   @Override
   public void channelActive(ChannelHandlerContext ctx) throws Exception {
-    xctx.getRequestMeter().mark();
+    ctx.channel().attr(CONNECTION_CONTEXT).get().getRequestMeter().mark();
   }
 
   private void executeHandler(ChannelHandlerContext ctx, int streamId, Route route)
       throws IOException {
+    XrpcConnectionContext xctx = ctx.channel().attr(CONNECTION_CONTEXT).get();
     FullHttpResponse h1Resp;
+
     Optional<ImmutableMap<XHttpMethod, Handler>> handlerMapOptional =
         xctx.getRoutes()
             .get()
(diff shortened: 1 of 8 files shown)


src/main/java/com/nordstrom/xrpc/server/Http2OrHttpHandler.java
```diff
@@ -7,22 +7,26 @@
 import io.netty.handler.codec.http.HttpServerCodec;
 import io.netty.handler.ssl.ApplicationProtocolNames;
 import io.netty.handler.ssl.ApplicationProtocolNegotiationHandler;
+import io.netty.util.AttributeKey;
 
 @ChannelHandler.Sharable
 public class Http2OrHttpHandler extends ApplicationProtocolNegotiationHandler {
-  private static final int MAX_PAYLOAD_SIZE =
-      1 * 1024 * 1024; //TODO(JR): This should be configurable
+  private static final AttributeKey<XrpcConnectionContext> CONNECTION_CONTEXT =
```
jkinkead: Might be worth putting the keys in a common (public) location, especially if the user is meant to use them to access these items.

I notice that this particular key is redefined in several places.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -70,6 +56,8 @@
   private final int bossThreadCount;
   private final int workerThreadCount;
   private final int MAX_PAYLOAD_SIZE;
+  //private static final AttributeKey<XioConnectionContext> CONNECTION_CONTEXT = AttributeKey.valueOf("XioConnectionContext")
```
jkinkead: Remove?

src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -12,20 +12,20 @@
...
 public class UrlRouter extends ChannelDuplexHandler {
-  private final XrpcChannelContext xctx;
+  private static final AttributeKey<XrpcConnectionContext> CONNECTION_CONTEXT =
+      AttributeKey.valueOf("XrpcConnectionContext");
 
-  public UrlRouter(XrpcChannelContext ctx) {
-
-    this.xctx = ctx;
-  }
+  public UrlRouter() {}
```
jkinkead: Superfluous, remove.

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -143,10 +152,13 @@ public void onHeadersRead(
...
     String uri = headers.path().toString();
-    for (Route route : xctx.getRoutes().get().descendingKeySet()) {
+    for (Route route :
+        ctx.channel().attr(CONNECTION_CONTEXT).get().getRoutes().get().descendingKeySet()) {
       Optional<Map<String, String>> groups = Optional.ofNullable(route.groups(uri));
       if (groups.isPresent()) {
-        xrpcRequest = new XrpcRequest(headers, groups.get(), ctx.channel(), streamId);
+        ctx.channel()
+            .attr(XRPC_REQUEST)
+            .set(new XrpcRequest(headers, groups.get(), ctx.channel(), streamId));
```
jkinkead: Why is this being updated? Is it to avoid the shared request member?

src/main/java/com/nordstrom/xrpc/server/Http2OrHttpHandler.java
pdex: Please don't expose the user to the `attr` bucket, instead make static getter/setter methods:
https://github.com/xjdr/xio/blob/master/src/main/java/com/xjeffrose/xio/SSL/TlsAuthState.java

pdex: I took a closer look and I don't think we're expecting the user to interact with the `attr` bucket, hence `private`. Please disregard my previous statement.

pdex:
I don't think this addresses #43 did you mean #47 ?

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
xjdr: Yes and to avoid creating a new object for each class and only passing down the single object instantiated at the onset of the request

xjdr:
I think this addresses #43 an #47

PR #51, opened 2017-12-24
Return 429 on too many requests and fix npe

xjdr:
This commit addresses issues 45 and 50.

For Issue 45:
2 new limits were added to the config, a "soft" limit and a "hard" limit.
Once the soft limit is reached, the server will return 429s directly. If the
hard limit is reached, the server will block on the worker thread to protect
the back end from being overwhelmed and thus leading to resource exhaustion.

For Issue 50:
The healthCheckRegistry was be scheduled to run prior to the worker group
being properly bootstrapped. This commit changes the order in which those
processes are run. This was added to this commit for convenience.

### DIFF
--- a/src/main/java/com/nordstrom/xrpc/XConfig.java
+++ b/src/main/java/com/nordstrom/xrpc/XConfig.java
@@ -35,7 +35,8 @@ public class XConfig {
   private final int bossThreadCount;
   private final int workerThreadCount;
   private final int maxConnections;
-  private final double rateLimit;
+  private final double softRateLimit;
+  private final double hardRateLimit;
   private final String cert;
   private final String key;
   private final int port;
@@ -64,7 +65,8 @@ public class XConfig {
     bossThreadCount = config.getInt("boss_thread_count");
     workerThreadCount = config.getInt("worker_thread_count");
     maxConnections = config.getInt("max_connections");
-    rateLimit = config.getDouble("req_per_sec");
+    softRateLimit = config.getDouble("soft_req_per_sec");
+    hardRateLimit = config.getDouble("hard_req_per_sec");
     cert = config.getString("cert");
     key = config.getString("key");
     port = config.getInt("server.port");
@@ -98,8 +100,12 @@ public class XConfig {
     return maxConnections;
   }
 
-  public double rateLimit() {
-    return rateLimit;
+  public double softRateLimit() {
+    return softRateLimit;
+  }
+
+  public double hardRateLimit() {
+    return hardRateLimit;
   }
 
   public String cert() {
--- a/src/main/java/com/nordstrom/xrpc/XConfig.java
+++ b/src/main/java/com/nordstrom/xrpc/XConfig.java
@@ -35,8 +35,8 @@ public class XConfig {
   private final int bossThreadCount;
   private final int workerThreadCount;
   private final int maxConnections;
-  private final double softRateLimit;
-  private final double hardRateLimit;
+  private final double softReqPerSec;
+  private final double hardReqPerSec;
   private final String cert;
   private final String key;
   private final int port;
@@ -65,8 +65,8 @@ public class XConfig {
     bossThreadCount = config.getInt("boss_thread_count");
     workerThreadCount = config.getInt("worker_thread_count");
     maxConnections = config.getInt("max_connections");
-    softRateLimit = config.getDouble("soft_req_per_sec");
-    hardRateLimit = config.getDouble("hard_req_per_sec");
+    softReqPerSec = config.getDouble("soft_req_per_sec");
+    hardReqPerSec = config.getDouble("hard_req_per_sec");
     cert = config.getString("cert");
     key = config.getString("key");
     port = config.getInt("server.port");
@@ -100,12 +100,12 @@ public class XConfig {
     return maxConnections;
   }
 
-  public double softRateLimit() {
-    return softRateLimit;
+  public double softReqPerSec() {
+    return softReqPerSec;
   }
 
-  public double hardRateLimit() {
-    return hardRateLimit;
+  public double hardReqPerSec() {
+    return hardReqPerSec;
   }
 
   public String cert() {
--- a/src/main/java/com/nordstrom/xrpc/server/ConnectionLimiter.java
+++ b/src/main/java/com/nordstrom/xrpc/server/ConnectionLimiter.java
@@ -27,10 +27,10 @@ public class ConnectionLimiter extends ChannelDuplexHandler {
   public void channelActive(ChannelHandlerContext ctx) throws Exception {
     connections.inc();
 
+    //TODO(JR): Should this return a 429 or is the current logic of silently dropping the connection sufficient?
     if (maxConnections > 0) {
       if (numConnections.incrementAndGet() > maxConnections) {
         ctx.channel().close();
-        // numConnections will be decremented in channelClosed
         log.info("Accepted connection above limit (" + maxConnections + "). Dropping.");
       }
     }
--- a/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
+++ b/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
@@ -162,7 +162,6 @@ public final class Http2Handler extends Http2ConnectionHandler implements Http2F
       int padding,
       boolean endOfStream) {
 
-    // This object is being called without an Optional<Boolean> to limit object creation and thus reduce GC pressure
     if (ctx.channel().hasAttr(XrpcConstants.XRPC_RATE_LIMIT)) {
       writeResponse(
           ctx,
--- a/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
+++ b/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
@@ -145,7 +145,7 @@ public final class Http2Handler extends Http2ConnectionHandler implements Http2F
             log.error("Error in handling Route", e);
             // Error
             ByteBuf buf = ctx.channel().alloc().directBuffer();
-            buf.writeBytes("Error executing endpoint".getBytes());
+            buf.writeBytes("Error executing endpoint".getBytes(XrpcConstants.DEFAULT_CHARSET));
             writeResponse(ctx, streamId, HttpResponseStatus.INTERNAL_SERVER_ERROR, buf);
           }
         }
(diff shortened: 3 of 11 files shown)


src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -64,7 +65,8 @@ public XConfig(Config configOverrides) {
     bossThreadCount = config.getInt("boss_thread_count");
     workerThreadCount = config.getInt("worker_thread_count");
     maxConnections = config.getInt("max_connections");
-    rateLimit = config.getDouble("req_per_sec");
+    softRateLimit = config.getDouble("soft_req_per_sec");
```
jkinkead: Style nitpick: I prefer to have the variables in the `.conf` file match the variables in `.java`. This makes it MUCH easier to figure out how to map the config variables to values read out of `XConfig` - you don't need the config file, constructor, and accessors open at the same time.

In this case, the java vars should be `softReqPerSec` and `hardReqPerSec`.

src/main/java/com/nordstrom/xrpc/server/ConnectionLimiter.java
```diff
@@ -27,10 +27,10 @@ public ConnectionLimiter(MetricRegistry metrics, int maxConnections) {
   public void channelActive(ChannelHandlerContext ctx) throws Exception {
     connections.inc();
 
+    //TODO(JR): Should this return a 429 or is the current logic of silently dropping the connection sufficient?
```
jkinkead: It sure leaves us more open for DoS if we try to answer these, but at low QPS, it's a lot more user-friendly to have a response...

I also don't love the idea of a hung service looking a lot like a down service or network error.

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -162,6 +162,20 @@ public void onHeadersRead(
       int padding,
       boolean endOfStream) {
 
+    // This object is being called without an Optional<Boolean> to limit object creation and thus reduce GC pressure
```
jkinkead: I don't think this uses `Optional`, which makes it confusing. You have an unset attribute, yes?

Does `hasAttr` work here? Also, if it doesn't work (if the `Attribute` is always set), you may as well set it to `false` in the rate limiting code.

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -162,6 +162,20 @@ public void onHeadersRead(
...
+    if (ctx.channel().attr(XrpcConstants.XRPC_RATE_LIMIT).get() != null) {
+      writeResponse(
+          ctx,
+          streamId,
+          HttpResponseStatus.TOO_MANY_REQUESTS,
+          ctx.alloc()
+              .directBuffer()
+              .writeBytes(
+                  "This respone is being send due to too many requests being sent to the server"
+                      .getBytes()));
```
jkinkead: These bytes should be in a static constant (and you should specify an encoding to `getBytes`).

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -119,14 +123,23 @@ private void configResponseCodeMeters() {
...
   }
 
-  public void scheduleHealthChecks() {
-    scheduleHealthChecks(60, 60, TimeUnit.SECONDS);
+  public void scheduleHealthChecks(EventLoopGroup workerGroup) {
+    scheduleHealthChecks(workerGroup, 60, 60, TimeUnit.SECONDS);
   }
 
-  public void scheduleHealthChecks(int initialDelay, int delay, TimeUnit timeUnit) {
+  public void scheduleHealthChecks(
```
jkinkead: You should note that this is not thread-safe (it sets a member variable) and should be called ONLY from the main thread (or better, when a lock is held).

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -6,32 +6,39 @@
...
   @Override
   public void channelActive(ChannelHandlerContext ctx) throws Exception {
-    // TODO(JR): Should this be before or after the acquire? Do we want to know when
-    //          we are limiting? Do we want to know what the actual rate of incoming
-    //          requests are?
     reqs.mark();
-    limiter.acquire();
-    context = timer.time();
 
+    hardLimiter.acquire();
```
jkinkead: So, I don't really know how Netty works, but this worries me. A single connection spamming a huge number of requests could theoretically overwhelm the server, which is really not-great.

Is there a way to drop the connection HERE, in addition to in the connection pool?

src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -24,6 +27,22 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
     XrpcConnectionContext xctx = ctx.channel().attr(XrpcConstants.CONNECTION_CONTEXT).get();
     xctx.getRequestMeter().mark();
 
+    // This object is being called without an Optional<Boolean> to limit object creation and thus reduce GC pressure
```
jkinkead: Same comment as above: Comment confusing; maybe use `hasAttr`?

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -6,32 +6,39 @@
...
+
+    if (!softLimiter.tryAcquire()) {
+      ctx.channel().attr(XrpcConstants.XRPC_RATE_LIMIT).set(true);
```
jkinkead: FWIW you'll avoid a method call if you replace `true` with `Boolean.TRUE`.

src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -24,6 +27,22 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
+    if (ctx.channel().attr(XrpcConstants.XRPC_RATE_LIMIT).get() != null) {
+      ctx.writeAndFlush(
+              Recipes.newResponse(
+                  HttpResponseStatus.TOO_MANY_REQUESTS,
+                  ctx.alloc()
+                      .directBuffer()
+                      .writeBytes(
+                          "This respone is being send due to too many requests being sent to the server"
+                              .getBytes()),
```
jkinkead: Some comment as above: Should be static; should specify encoding.

src/main/java/com/nordstrom/xrpc/server/Router.java
xjdr: This was an oversight. moving that member variable creation to the listenAndServe() method after that worker group has been bootstrapped. I add a note to listenAndServe() method about only being called once from the main thread.

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
xjdr: Updated to use hasArrt()

src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
xjdr: Updated to use hasArrt()

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
xjdr: acceptor thread pool accepts requests and then dispatches them to a worker pool -> worker pool is where the channel pipeline is executed. The code above will block one of the workers for as much time as it takes to be able to acquire a slice of the hardLimiter. This helps performance and makes the system more deterministic but leave the system more exposed to DOS. That said, the connection limited should be able to make up for this as it is ahead of the rate limited in the pipeline. 

Opened https://github.com/Nordstrom/xrpc/issues/55 to address directly.

src/main/java/com/nordstrom/xrpc/server/ConnectionLimiter.java
xjdr: Opened issue #56 to address

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -162,6 +162,19 @@ public void onHeadersRead(
...
+    if (ctx.channel().hasAttr(XrpcConstants.XRPC_RATE_LIMIT)) {
+      writeResponse(
+          ctx,
+          streamId,
+          HttpResponseStatus.TOO_MANY_REQUESTS,
+          ctx.alloc()
+              .directBuffer()
+              .writeBytes(
+                  "This respone is being send due to too many requests being sent to the server"
+                      .getBytes(XrpcConstants.DEFAULT_CHARSET)));
```
jkinkead: This is still creating a new byte array with each invocation, which isn't great for a rate limiter.

You could probably create an entirely static buffer, too, which would be even better under DoS conditions.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -210,13 +224,26 @@ public void serveAdmin() {
...
 
   public void listenAndServe() throws IOException {
+    listenAndServe(true, true);
+  }
+
+  /**
+   * The listenAndServe method is the primary entry point for the server and should only be called
+   * once and only from the main thread.
+   *
+   * @param serveAdmin
```
jkinkead: Should complete the `@param / @throws` tags or remove.

src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -24,6 +27,21 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
 
+    if (ctx.channel().hasAttr(XrpcConstants.XRPC_RATE_LIMIT)) {
+      ctx.writeAndFlush(
+              Recipes.newResponse(
+                  HttpResponseStatus.TOO_MANY_REQUESTS,
+                  ctx.alloc()
+                      .directBuffer()
+                      .writeBytes(
+                          "This respone is being send due to too many requests being sent to the server"
+                              .getBytes(XrpcConstants.DEFAULT_CHARSET)),
```
jkinkead: Consider constant response bytes here, too.

Issue #48, opened 2017-12-21
Path Variables: final variable includes query string


forplayplay:
for example, for the path:
`/people/{person}`

the following uri
`/people/patrick?exampleParam=something`

produces the following context:
`{person: "patrick?exampleParam=something"}`

we expect that it would be:
`{person: "patrick"}`

jkinkead:
This looks like a duplicate of #27 . Are you using the latest xrpc?

forplayplay:
It seemed as though #27 was fixed in 72a47e635f4c3469a0b9be810d1c891b17facf75 but it only fixed this issue for http1 requests.

xjdr:
Addressed by 02fa63f918e4a34571a1541811a4e3ba23208f8f

Issue #27, opened 2017-12-04
Router should match on path ONLY, not full URI


jkinkead:
[This line](https://github.com/Nordstrom/xrpc/blob/83a24cb6739d5273a7b99923fc0aa05e4809a513/src/main/java/com/nordstrom/xrpc/server/UrlRouter.java#L39) should be matching on the path ONLY. This matches query parameters, since it hits the whole URI.

shine17:
I thinks this needs to be handled for both http 1.1 and http2 handlers

jkinkead:
Patching `UrlRouter` should fix behavior in both HTTP/1.x & HTTP/2.

xjdr:
solved by 1eb5226ef5a3c4bb8b34462353b51f02bca8cb98

PR #54, opened 2017-12-26
issue 48 - query params in path variable

forplayplay:
https://github.com/Nordstrom/xrpc/issues/48

The previous fix for this problem (https://github.com/Nordstrom/xrpc/issues/27) left http2 unresolved. This takes applies the same solution to http2 as was used for http1.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,6 +69,7 @@ dependencies {
     testCompile 'org.junit.jupiter:junit-jupiter-api:5.0.1'
     testCompile 'org.apiguardian:apiguardian-api:1.0.0'
     testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.0.1'
+    testCompile 'org.mockito:mockito-core:2.13.0'
 }
 
 // Run with the latest checkstyle version. Required for our checkstyle.xml.
--- a/build.gradle
+++ b/build.gradle
@@ -70,8 +70,8 @@ dependencies {
     compile 'com.github.spotbugs:spotbugs:3.1.0'
     testCompile 'org.junit.jupiter:junit-jupiter-api:5.0.1'
     testCompile 'org.apiguardian:apiguardian-api:1.0.0'
-    testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.0.1'
     testCompile 'org.mockito:mockito-core:2.13.0'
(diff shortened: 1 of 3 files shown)


xjdr:
If you can write a unit test to demonstrate functionality, feel free to merge it afterward. Also, congrats @forplayplay on your first accepted PR!

build.gradle
```diff
@@ -71,6 +71,7 @@ dependencies {
     testCompile 'org.junit.jupiter:junit-jupiter-api:5.0.1'
     testCompile 'org.apiguardian:apiguardian-api:1.0.0'
     testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.0.1'
+    testCompile 'org.mockito:mockito-core:2.13.0'
```
jkinkead: Sort into `testCompile` section?

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -205,6 +207,11 @@ public void onHeadersRead(
     writeResponse(ctx, streamId, HttpResponseStatus.NOT_FOUND, buf);
   }
 
+  protected String getPathFromHeaders (Http2Headers headers) {
```
jkinkead: Should be default access; you could also add a [`@VisibleForTesting`](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/annotations/VisibleForTesting.html) annotation if you're feeling like it.

You could also make this `static`, which would make testing easier.

src/test/java/com/nordstrom/xrpc/server/Http2HandlerTest.java
```diff
@@ -0,0 +1,26 @@
...
+import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.handler.codec.http2.Http2Settings;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import org.junit.jupiter.api.Test;
+
+
+public class Http2HandlerTest {
+  @Test
+  public void getPathFromHeaders () {
```
jkinkead: I think this is fine, but it doesn't really test your fix per se, since the key is that this wasn't being used to parse groups.

Anyway, I don't think you need to burn more time on this right now.

Maybe add a test or two with queryless paths?

jkinkead:
LGTM, I'll go ahead and merge.

PR #58, opened 2017-12-27
Fix spotbugs issues -

xjdr:
Various commits to address deprecating infer in favor of spotbugs (for now) and addressing the all of the new issues brought up by that tool.


src/main/java/com/nordstrom/xrpc/server/Firewall/XrpcFirewall.java
```diff
@@ -52,30 +53,31 @@ private void buildReqCtx(ChannelHandlerContext ctx) {
...
-        log.info("Xrpc Firewall blocked based on policy:" + ctx.channel());
-        ctx.channel().deregister();
-        break;
-      case BLOCK_REQ_BEHAVIORAL_BASED:
-        log.info("Xrpc Firewall blocked based on behavior:" + ctx.channel());
-        ctx.channel().deregister();
-        break;
-      default:
-        break;
+    if (evt.isPresent()) {
```
jkinkead: Or monod-style with lambda:
```java
evt.ifPresent(evt -> {
  switch(evt) {
  }
});
```
This requires `ctx` to be `final`, but that's fine.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -319,22 +322,107 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
                 synchronized (Router.this) {
-                  // listener.serverShutdown();
+                  //TODO(JR): We should probably be more thoughtful here.
+                  shutdown();
                 }
               }
             });
   }
 
+  /**
```
jkinkead: Should be non-javadoc multiline (this will be bound to just `ConnectionLimiter` as-is).

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -319,22 +322,107 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
+   * This section of classes have been added as Static inner classes to reduce GC pressure. As these
```
jkinkead: I'm afraid I don't understand the GC pressure point. Inner classes makes sense from scoping (your point below), but what does that have to do with GC?

src/main/java/com/nordstrom/xrpc/server/tls/Tls.java
```diff
@@ -79,14 +79,14 @@ public SslContext buildEncryptionHandler() {
       final List<java.security.cert.X509Certificate> certList = new ArrayList<>();
       final String rawCertString = cert;
       PrivateKey privateKey;
-      PublicKey publicKey;
+      //PublicKey publicKey;
```
jkinkead: Remove commented out code?

PR #59, opened 2017-12-27
Added another newResponse method to address issue 32. 

lducharme:
New method takes a map of strings that allows the caller to specify custom headers.

Also, please feel free to let me know if I've misunderstood anything about this fix or the PR process. Thanks!


src/main/java/com/nordstrom/xrpc/server/http/Recipes.java
```diff
@@ -107,6 +107,27 @@ public static FullHttpResponse newResponse(
     return response;
   }
 
+  /*
```
jkinkead: Should be javadoc formatted (`/**` at start).

src/main/java/com/nordstrom/xrpc/server/http/Recipes.java
```diff
@@ -107,6 +107,27 @@ public static FullHttpResponse newResponse(
...
+   * Returns a full HTTP response with the specified status, content type, and custom headers.
+   *
+   * Headers should be specified as a map of strings. For example, to allow CORS, add the
+   * following key and value:
+   *     "access-control-allow-origin", "*"
+   */
+  public static FullHttpResponse newResponse(
+    HttpResponseStatus status, ByteBuf buffer, ContentType contentType, Map<String, String> customHeaders) {
```
jkinkead: There's a mix of the request content argument being named `buffer` and being named `payload` in this file. Would you mind renaming all of them to `payload`, for consistency & clarity?

src/main/java/com/nordstrom/xrpc/server/http/Recipes.java
```diff
@@ -107,6 +107,27 @@ public static FullHttpResponse newResponse(
     return response;
   }
 
+  /*
+   * Returns a full HTTP response with the specified status, content type, and custom headers.
+   *
+   * Headers should be specified as a map of strings. For example, to allow CORS, add the
+   * following key and value:
+   *     "access-control-allow-origin", "*"
```
jkinkead: We never want an allowed origin of `*`.

A better method to handle CORS is to have a configured set of allowed origins (possibly with wildcards), and echo back the request's `Origin` header as the allowed origin when it matches the configuration.

src/main/java/com/nordstrom/xrpc/server/http/Recipes.java
```diff
@@ -107,6 +107,27 @@ public static FullHttpResponse newResponse(
...
+   */
+  public static FullHttpResponse newResponse(
+    HttpResponseStatus status, ByteBuf buffer, ContentType contentType, Map<String, String> customHeaders) {
+    FullHttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status, buffer);
+
+    response.headers().set(CONTENT_TYPE, contentType.value);
```
jkinkead: These should probably go after the header iteration.

Either way, you should document the behavior (`customHeaders` overrides provided content type + length, or `customHeaders` content type + length are ignored).

lducharme:
Jesse, thanks for the comments! I think I've addressed most of them with the last few commits.

Hmm, that's a good point about there being a better way to handle CORS- this change really is for custom headers, not CORS specifically. I will take a look at implementing that today.

src/main/java/com/nordstrom/xrpc/server/http/Recipes.java
```diff
@@ -98,11 +99,36 @@ public static FullHttpResponse newResponse(HttpResponseStatus status) {
   }
 
   public static FullHttpResponse newResponse(
-      HttpResponseStatus status, ByteBuf buffer, ContentType contentType) {
-    FullHttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status, buffer);
+      HttpResponseStatus status, ByteBuf payload, ContentType contentType) {
+    FullHttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status, payload);
```
jkinkead: Update this to use your new method?

```java
// java.util.Maps below.
return newResponse(status, payload, contentType, Maps.emptyMap());
```

lducharme:
I fixed the conflict that this request had- what is the next step to merge? @jkinkead

jkinkead:
I will merge now. :)

lducharme:
I think this change needs to be released before I can pull it in as a dependency in another project, right? If I follow the instructions in [CONTRIBUTING.md](https://github.com/Nordstrom/xrpc/blob/master/CONTRIBUTING.md), can I go ahead and do this?

jkinkead:
If you'd like, yes. Else, either @spenserpothier or myself can run a release.

PR #61, opened 2017-12-28
RPC demo

xjdr:
[WIP] - first pass at RPC demo. Leaving here for comment. 

Working now. Good for review

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,93 +1,10 @@
 xrpc
 ====
 
-Simple, production ready Java API server built on top of functional composition.
-
-What you get:
-1) Openssl Native TLS w/ direct x509 pem support (vs keytool)
-2) Epoll native
-3) RateLimiting
-4) Connection Limiting
-5) Log forwarder Integration
-6) Metrics & Logs
-7) Binary Debug for packet level inspection
-8) Protobuf support
-9) Docker
-10) k8s
-11) Many Much more
-
-Make .jar not .war ...
-
-Example.java
-```java
-@Slf4j
-public class Example {
-  /** Example POJO for use in request / response. */
-  @AllArgsConstructor
-  private static class Person {
-    private String name;
-  }
-
-  public static void main(String[] args) {
-    final List<Person> people = new ArrayList<>();
-
-    // See https://github.com/square/moshi for the Moshi Magic.
-    Moshi moshi = new Moshi.Builder().build();
-    Type type = Types.newParameterizedType(List.class, Person.class);
-    JsonAdapter<List<Person>> adapter = moshi.adapter(type);
-
-    // Load application config from jar resources. The 'load' method below also allows supports
-    // overrides from environment variables.
-    Config config = ConfigFactory.load("demo.conf");
-
-    // Build your router. This overrides the default configuration with values from
-    // src/main/resources/demo.conf.
-    XConfig xConfig = new XConfig(config.getConfig("xrpc"));
-    Router router = new Router(xConfig);
-
-    // Define a simple function call.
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -1,93 +1,16 @@
...
-      router.listenAndServe();
-    } catch (IOException e) {
-      log.error("Failed to start people server", e);
-    }
-  }
-}
- 
-```
+Xrpc is a framework for creating production quality API services on top of Netty. The framework helps to encapsulate our
+best practices and provides sane (ans safe) defaults.
```
jkinkead: s/ans/and/

README.md
```diff
@@ -1,93 +1,16 @@
...
+
+It currently supports the http/1.1 and the http/2 protocols. It does so interchangeably, 
+i.e your implementation does not need to change and it will automatically respond to a http/1.1 
+and a http/2 client the same way. The user is free to determine whatever payload they would like, 
+but our recommendation is JSON where you don't control both ends and protobuf (version 3) where you do
```
jkinkead: Missing terminal period.

README.md
```diff
@@ -1,93 +1,16 @@
...
 
+## Testing with the Example class
```
jkinkead: Remove one `#` here to match other headers?

demo/src/main/java/com/nordstrom/xrpc/demo/Example.java
```diff
@@ -59,94 +67,54 @@ public static void main(String[] args) {
...
               Recipes.ContentType.Application_Json);
         };
 
     // Define a complex function call
-    Handler personHandler =
-        context -> {
-          Person p = new Person(context.variable("person"));
+    Handler personPostHandler =
+        request -> {
+          byte[] _p = new byte[request.getData().readableBytes()];
```
jkinkead: `buf` or `data` instead of `_p`?

demo/src/main/proto/dino.proto
```diff
@@ -21,6 +21,35 @@ package nordstrom.dino;
 option java_package = "com.nordstrom.xrpc.demo.proto";
 option java_multiple_files = true;
 
+// The greeting service definition.
```
jkinkead: Update? This looks like a copy-paste description.

demo/src/main/java/com/nordstrom/xrpc/demo/Example.java
```diff
@@ -17,33 +17,41 @@
...
 @Slf4j
 public class Example {
   public static void main(String[] args) {
     final List<Person> people = new ArrayList<>();
-    //    final List<Dino> dinos = new ArrayList<>();
+    final List<Dino> dinos = new ArrayList<>();
 
-    // See https://github.com/square/moshi for the Moshi Magic.
+    // For the JSON portion of the demo
     Moshi moshi = new Moshi.Builder().build();
```
jengc: one question, if we go from bytebuf to moshi back to bytebuf, don't we through assignment/init incur the overhead of gc'd strings versus having a custom bytebuf json client?

jengc: not that it's a blocker, but if it's our primary payload type for api's, it may be worth investigating.

xjdr: There are several efficiencies we could make here, this is more for demonstration purposes. Essentially, we are only ever working with bytes and their wrappers. For a latency / memory sensitive application, i would recommend protobuf or something more custom that more closely controlled the memory allocation

Issue #41, opened 2017-12-14
XUrl should be a wrapper around okhttp3.HttpUrl


pdex:
https://github.com/Nordstrom/xrpc/pull/38#discussion_r157033917

PR #65, opened 2017-12-30
Issue 44

forplayplay:
Fixes issue 44 by having XUrl do percent decoding before returning values to the client.

- Refactors XUrl to remove the java.net.URL class which was used heavily and inefficient (methods using it were sometimes taking milliseconds). Removes methods that handles checked exceptions from java.net.URL.
- Removes duplicate/unused methods.
- Adopts consistent naming strategy for methods.
- Implements a new class with an empty default for query parameters for more graceful failures when clients attempt to read query params that aren't present.
- Minor change to demo App for easier manual testing.


src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -16,71 +16,72 @@
...
 
   public static String getHost(String url) {
-    try {
-      return getDomainChecked(url);
-    } catch (URISyntaxException e) {
-      log.info("Malformed url: " + url);
-      return null;
+    Preconditions.checkNotNull(url);
+
+    QueryStringDecoder decoder = new QueryStringDecoder(url);
```
jkinkead: A few notes:
- any decoding MUST happen after query / path / host / schema parsing
- this isn't meant to parse the host, and probably won't work

You should be using [HttpUrl](https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.html). It might be worth just fixing #41 .

PR #67, opened 2018-01-03
Limit per client 

xjdr:
This PR sets up the basic framework for rate limiting on a per client basis as well as ip white listing and ip black listing. Still to add is the ability to allow special rate limits per endpoint and to implement the ip whitelist. 

UPDATE:
{White,Black} List impl added


src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -40,6 +42,10 @@
   private final String cert;
   private final String key;
   private final int port;
+  private final double gloablSoftReqPerSec;
+  private final double globalHardReqPerSec;
+  private final List<String> ipBlackList;
+  private final List<String> ipWhiteList;
```
jkinkead: Make [`ImmutableSet`](http://google.github.io/guava/releases/23.6-jre/api/docs/com/google/common/collect/ImmutableSet.html), and build in constructor?

src/main/java/com/nordstrom/xrpc/XrpcConstants.java
```diff
@@ -18,4 +21,8 @@
           .writeBytes(
               "This response is being send due to too many requests being sent to the server"
                   .getBytes(DEFAULT_CHARSET));
+  public static final AttributeKey<? extends Object> IP_WHITE_LIST =
```
jkinkead: Should this match the other limits and use `Boolean`?

src/main/java/com/nordstrom/xrpc/server/ConnectionLimiter.java
```diff
@@ -0,0 +1,67 @@
...
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.MetricRegistry;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import java.util.concurrent.atomic.AtomicInteger;
+import lombok.extern.slf4j.Slf4j;
+
+/*
+ * This section of classes have been added as Static inner classes to reduce GC pressure. As these
```
jkinkead: Out-of-date?

src/main/java/com/nordstrom/xrpc/server/ConnectionLimiter.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Override
+  public void channelActive(ChannelHandlerContext ctx) throws Exception {
+    connections.inc();
+
+    //TODO(JR): Should this return a 429 or is the current logic of silently dropping the connection sufficient?
+    if (maxConnections > 0) {
+      if (numConnections.incrementAndGet() > maxConnections) {
+        log.info("Accepted connection above limit (" + maxConnections + "). Dropping.");
```
jkinkead: Should use format string + multiple arguments to avoid formatting if info is disabled, and to avoid `StringBuilder`:
```java
log.info("Accepted connection above limit (%d). Dropping.", maxConnections);
```

src/main/java/com/nordstrom/xrpc/server/Firewall.java
```diff
@@ -0,0 +1,55 @@
...
+  public Firewall(MetricRegistry metrics) {
+
+    this.metrics = metrics;
+    this.rateLimits = metrics.meter("Hard Rate Limits");
+  }
+
+  @Override
+  public void channelActive(ChannelHandlerContext ctx) throws Exception {
+    if (ctx.channel().hasAttr(XrpcConstants.XRPC_HARD_RATE_LIMIT)) {
+      log.debug("Channel " + ctx.channel() + " Closed due to Xrpc Hard Rate Limit being reached");
```
jkinkead: Use multi-arg `debug` with format string.

src/main/java/com/nordstrom/xrpc/server/Firewall.java
```diff
@@ -0,0 +1,55 @@
...
+      rateLimits.mark();
+      ctx.pipeline().channel().closeFuture();
+    }
+
+    if ((ctx.channel().hasAttr(XrpcConstants.IP_BLACK_LIST))) {
+      log.debug("Channel " + ctx.channel() + " Closed due to Xrpc IP Black List Configuration");
```
jkinkead: Use multi-arg `debug` with format string.

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -80,6 +88,9 @@ public XConfig(Config configOverrides) {
     consoleReporter = config.getBoolean("console_reporter");
     slf4jReporterPollingRate = config.getInt("slf4j_reporter_polling_rate");
     consoleReporterPollingRate = config.getInt("console_reporter_polling_rate");
+
+    ipBlackList = config.getStringList("ip_black_list");
+    ipWhiteList = config.getStringList("ip_black_list");
```
jkinkead: Incorrect value being read (`"ip_black_list"` instead of `"ip_white_list"`).

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -149,4 +160,20 @@ public long slf4jReporterPollingRate() {
...
+
+  public double globalHardReqPerSec() {
+    return globalHardReqPerSec;
+  }
+
+  public double globalSoftReqPerSec() {
+    return gloablSoftReqPerSec;
+  }
+
+  public HashSet<String> ipBlacklist() {
```
jkinkead: Nit: Getters should match casing of members (second `l` should be capitalized, here and in white list accessor).

src/main/java/com/nordstrom/xrpc/server/WhiteListFilter.java
```diff
@@ -0,0 +1,35 @@
...
+    this.metrics = metrics;
+    this.whiteList = whiteList;
+  }
+
+  @Override
+  public void channelActive(ChannelHandlerContext ctx) throws Exception {
+    String remoteAddress = ((InetSocketAddress) ctx.channel().remoteAddress()).getHostName();
+
+    if (!whiteList.contains(remoteAddress)) {
+      ctx.channel().attr(XrpcConstants.IP_WHITE_LIST).set(Boolean.TRUE);
```
jkinkead: I would reverse this logic, both for readability ("true" implies not in whitelist is confusing) and for efficiency under DoS (not adding an attribute to all channels coming from unknown sources).

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -273,6 +277,9 @@ public void initChannel(Channel ch) throws Exception {
 
     ChannelFuture future = b.bind(new InetSocketAddress(config.port()));
 
+    // Clean the rate limiter map so we do not introduce a memory leak
```
jkinkead: I feel like this is a crufty way to keep the list from growing unbounded - and it wouldn't work if someone were DoSing with a large number of IPs, since they'd blow the heap before this 5-minute window elapsed.

What about using a fixed-size rate limiter pool, and hash the IP address to fetch the limiter for that IP? This would lose the extra garbage per request, and solve the unbounded growth problem as well.

Dockerfile
```diff
@@ -1,18 +1,53 @@
-FROM openjdk:8-jre-alpine
+## -*- docker-image-name: "nordstrom/xrpc" -*-
+FROM debian:stretch
```
jkinkead: Not Ubuntu?

Dockerfile
```diff
@@ -1,18 +1,53 @@
...
+
+# ===== Installing packages =====
+RUN apt-get install -y \
+    git-core \
+    wget \
+    unzip \
+    curl \
+    libssl-dev \
+    openjdk-8-jdk \
+--no-install-recommends
```
jkinkead: Move to top & indent?

Dockerfile
```diff
@@ -1,18 +1,53 @@
...
-RUN apk --update upgrade && \
-    # install JDK
-    apk add curl openjdk8="$JAVA_ALPINE_VERSION" && \
-    # build the application into a single JAR, including dependencies
-    ./gradlew shadowJar && \
-    mv build/libs/*-all.jar app.jar
-
-# TODO(jkinkead): Fix this to use a base image instead of keeping around the build tools.
+RUN ./gradlew clean shadowJar && \
+  mv demo/build/libs/xrpc-demo-0.1.1-SNAPSHOT-all.jar app.jar
```
jkinkead: TODO to use a multi-stage build to copy the shadow jar into an execution environment?

Dockerfile
```diff
@@ -1,18 +1,53 @@
...
+RUN apt-get install -y \
+    git-core \
+    wget \
+    unzip \
+    curl \
+    libssl-dev \
+    openjdk-8-jdk \
+--no-install-recommends
+
+# ===== Build xrpc  =====
```
jkinkead: Duplicated on lines 47-50.

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -149,4 +194,32 @@ public long slf4jReporterPollingRate() {
   public long consoleReporterPollingRate() {
     return (long) consoleReporterPollingRate;
   }
+
+  public double globalHardReqPerSec() {
```
jkinkead: Use lombok `@Getter` for these instead?

src/main/java/com/nordstrom/xrpc/XrpcConstants.java
```diff
@@ -9,7 +9,10 @@
 
 public class XrpcConstants {
   public static final AttributeKey<XrpcRequest> XRPC_REQUEST = AttributeKey.valueOf("XrpcRequest");
-  public static final AttributeKey<Boolean> XRPC_RATE_LIMIT = AttributeKey.valueOf("XrpcRateLimit");
+  public static final AttributeKey<Boolean> XRPC_SOFT_RATE_LIMITED =
+      AttributeKey.valueOf("XrpcSoftRateLimit");
```
jkinkead: Update string to match new name, here and below?

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -0,0 +1,136 @@
...
+    for (int i = 0; i < poolSize; i++) {
+      String id = UUID.randomUUID().toString();
+      softTempPool.add(id);
+      softLimiterMap.put(id, RateLimiter.create(config.softReqPerSec()));
+    }
+
+    return new RendezvousHash(Funnels.stringFunnel(XrpcConstants.DEFAULT_CHARSET), softTempPool, 1);
+  }
+
+  private RendezvousHash buildHardHasher(int poolSize) {
```
jkinkead: The two hash builder functions are identical except for the config request-per-sec value read. Make them a single function with two parameters instead of two functions with one parameter?

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@ChannelHandler.Sharable
+class ServiceRateLimiter extends ChannelDuplexHandler {
+  private final Map<String, RateLimiter> softLimiterMap = PlatformDependent.newConcurrentHashMap();
+  private final Map<String, RateLimiter> hardLimiterMap = PlatformDependent.newConcurrentHashMap();
```
jkinkead: I was picturing something more like:

```java

import java.util.RandomAccess;

// TODO: In constructor, initialize with com.google.common.collect.ImmutableList.
private final RandomAccess<RateLimiter> rateLimiters;

// When accessing:
RateLimiter limiter = rateLimiters.get(remodeAddress.hashCode() % limiterCount);
```
`ConcurrentHashMap` imposes a synchronization cost that the `Immutable*` collections don't. Also, you're (possibly?) creating new strings, hashing them, etc.

src/main/java/com/nordstrom/xrpc/RendezvousHash.java
```diff
@@ -0,0 +1,66 @@
...
+    return true;
+  }
+
+  boolean add(N node) {
+    nodeList = new ImmutableSet.Builder<N>().addAll(nodeList).add(node).build();
+    return true;
+  }
+
+  public List<N> get(byte[] key) {
+    List<N> _nodeList = new ArrayList<>();
```
jkinkead: This is creating a new list with each call? That seems not ideal.

src/main/java/com/nordstrom/xrpc/RendezvousHash.java
```diff
@@ -0,0 +1,66 @@
...
+            xs -> {
+              hashMap.put(
+                  hasher.newHasher().putBytes(key).putObject(xs, nodeFunnel).hash().asLong(), xs);
+            });
+
+    for (int i = 0; i < listSize; i++) {
+      _nodeList.add(hashMap.remove(hashMap.keySet().stream().max(Long::compare).orElse(null)));
+    }
+
+    hashMap.clear();
```
jkinkead: This doesn't look correct.

src/main/java/com/nordstrom/xrpc/RendezvousHash.java
```diff
@@ -0,0 +1,66 @@
...
+    nodeList = new ImmutableSet.Builder<N>().addAll(nodeList).add(node).build();
+    return true;
+  }
+
+  public List<N> get(byte[] key) {
+    List<N> _nodeList = new ArrayList<>();
+
+    nodeList
+        .parallelStream()
+        .filter(xs -> !_nodeList.contains(xs))
```
jkinkead: no-op, because `_nodeList` is always empty here?

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -0,0 +1,136 @@
...
+  private Map<ChannelHandlerContext, Timer.Context> timerMap = PlatformDependent.newConcurrentHashMap();
+
+  public ServiceRateLimiter(MetricRegistry metrics, XConfig config) {
+    this.reqs = metrics.meter(name(Router.class, "requests", "Rate"));
+    this.timer = metrics.timer("Request Latency");
+    this.config = config;
+    this.globalHardLimiter = RateLimiter.create(config.globalHardReqPerSec());
+    this.globalSoftLimiter = RateLimiter.create(config.globalSoftReqPerSec());
+
+    softRateLimitHasher = buildSoftHasher(config.getRateLimiterPoolSize());
```
jkinkead: A rendezvous hasher seems wrong given that the pool size isn't expected to change at runtime (no nodes being added / removed).

Dockerfile
xjdr: Debian is standard

Dockerfile
xjdr: This is more for testing & examples for the demo app than for deployment work. It should probably be moved to the `demo/` directory accordingly.

src/main/java/com/nordstrom/xrpc/RendezvousHash.java
```diff
@@ -0,0 +1,67 @@
...
+    hashMap.clear();
+
+    nodeList.forEach(
+        xs -> {
+          hashMap.put(
+              hasher.newHasher().putBytes(key).putObject(xs, nodeFunnel).hash().asLong(), xs);
+        });
+
+    for (int i = 0; i < listSize; i++) {
+      _nodeList.add(i, hashMap.remove(hashMap.keySet().stream().max(Long::compare).orElse(null)));
```
jkinkead: This is O(n^2) in `listSize`. Use `ConcurrentSkipListMap` instead - or if using a `ThreadLocal` to store the map, `TreeMap` or another `SortedMap`?

You could also just sort the key set first:
```java
SortedSet<Long> sortedKeys = new TreeSet(hashMap.keySet());
sortedKeys.descendingIterator().forEachRemaining(xs -> _nodeList.add(hashMap.get(xs));
```

src/main/java/com/nordstrom/xrpc/RendezvousHash.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    return Lists.newArrayList(_nodeList);
```
jkinkead: If you're creating a new array list of `listSize` anyway, there's no point in the class member `_nodeList`. You might as well create a local:

```java
List<N> _nodeList = new ArrayList<>(listSize);
```

demo/Dockerfile
```diff
@@ -0,0 +1,75 @@
+## -*- docker-image-name: "nordstrom/xrpc" -*-
+FROM debian:stretch
```
jkinkead: The reason I asked about debian is that I think we recommend ubuntu internally...

src/main/java/com/nordstrom/xrpc/RendezvousHash.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public void refresh(List<N> list) {
```
jkinkead: Same comment on `add` / `remove` - this isn't thread-safe.

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -0,0 +1,126 @@
...
+    return new RendezvousHash(Funnels.stringFunnel(XrpcConstants.DEFAULT_CHARSET), _tempPool, 1);
+  }
+
+  @Override
+  public void channelActive(ChannelHandlerContext ctx) throws Exception {
+    reqs.mark();
+
+    // Rate Limit per server
+    String remoteAddress =
+        ((InetSocketAddress) ctx.channel().remoteAddress()).getAddress().getHostAddress();
```
jkinkead: Jason pointed out to me that this won't work when sitting behind a proxy. We need to be looking at `X-Forwarded-For` values as well.

Maybe a TODO?

src/main/java/com/nordstrom/xrpc/RendezvousHash.java
```diff
@@ -0,0 +1,67 @@
...
+    nodeList = Sets.difference(nodeList, ImmutableSet.of(node)).immutableCopy();
+    return true;
+  }
+
+  boolean add(N node) {
+    nodeList = new ImmutableSet.Builder<N>().addAll(nodeList).add(node).build();
+    return true;
+  }
+
+  public List<N> get(byte[] key) {
```
jkinkead: Is there a use-case for using the non-first value returned from this function? If there isn't, it would be better to have this return a single value and avoid list creation.

If there is, we could avoid an extra list creation operation if we did something like:

```java
private SortedSet<Long> getKeys(byte[] key) {
  // TODO: Most of the logic; return the sorted map keys returned by the hasher.
}

public List<N> get(byte[] key) {
  SortedSet<Long> sortedKeys = getKeys(key);
  // TODO: Build sorted list.
}

public N get(byte[] key) {
  return hashMap.get(getKeys(key).first());
}
```

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
xjdr: Captured in https://github.com/Nordstrom/xrpc/issues/73

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -0,0 +1,93 @@
...
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+import com.google.common.util.concurrent.RateLimiter;
+import com.nordstrom.xrpc.XConfig;
+import com.nordstrom.xrpc.XrpcConstants;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.util.internal.PlatformDependent;
```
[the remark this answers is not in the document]
andyday: not sure its a good idea to use internal API calls.  also this will only select the V8 hashmap if java version < 8.  aren't we shooting for support of java 8+ here?

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
andyday: if this is the case, there is no need for rendevous hash since the hashcode will be moded by a limiterCount that can be inc'd/dec'd...  or am I missing something...

andyday: on the other hand this would change where ip limiters would end up in the hashtable based on a changing limiterCount...

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -0,0 +1,124 @@
...
+      log.debug("Global Hard Rate limit fired");
+      ctx.channel().attr(XrpcConstants.XRPC_HARD_RATE_LIMITED).set(Boolean.TRUE);
+
+    } else if (!globalSoftLimiter.tryAcquire()) {
+      ctx.channel().attr(XrpcConstants.XRPC_SOFT_RATE_LIMITED).set(Boolean.TRUE);
+    }
+
+    ctx.fireChannelActive();
+
+    timerMap.put(ctx, timer.time());
```
andyday: its not obvious to me where/how this is used.

demo/src/main/resources/demo.conf
```diff
@@ -4,6 +4,15 @@ mykey = "value"
 # Place any xrpc configuration overrides here.
 xrpc {
   worker_name_format = "demo-server-worker-%d"
-  req_per_sec = 500.0
+  soft_req_per_sec = 500.0
```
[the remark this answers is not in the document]
xjdr: 429 is returned

src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -22,14 +22,29 @@ worker_thread_count = 40
...
 # Set to zero to disable connection limiting.
-max_connections = 500
+max_connections = 2000
+
 # The maximum number of requests per second to allow before limiting new connection.
 # See: http://google.github.io/guava/releases/23.0/api/docs/com/google/common/util/concurrent/RateLimiter.html
 # Once the soft limit is hit, the server will return a HTTP 429 TOO MANY REQUESTS
 # If the hard limit is reached, the system will block to protect the back end resources
+
+# This is the per server rate limiter
```
jkinkead: This got injected between the rate-limit comment and the rate-limit settings. It should probably be below them.

src/main/java/com/nordstrom/xrpc/server/Firewall.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    if ((ctx.channel().hasAttr(XrpcConstants.IP_BLACK_LIST))) {
+      if (log.isErrorEnabled()) {
+        log.error("Channel " + ctx.channel() + " Closed due to Xrpc IP Black List Configuration");
+      }
+      ctx.channel().closeFuture();
+    }
+
+    if ((ctx.channel().hasAttr(XrpcConstants.IP_WHITE_LIST))) {
```
jkinkead: A comment (`This is always set to false`) would be really helpful here.

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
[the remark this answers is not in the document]
jkinkead: You should check out [the tls package](https://github.com/Nordstrom/xrpc/tree/master/src/main/java/com/nordstrom/xrpc/server/tls) if you're worried about internal libraries ... :)

This is only used for a hash map, so I wouldn't worry about the cost of porting, if it's removed or whatever.

demo/src/main/resources/demo.conf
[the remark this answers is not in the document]
jkinkead: This is documented in the main config, too.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -202,14 +198,29 @@ public void serveAdmin() {
...
         AdminHandlers.healthCheckHandler(healthCheckRegistry, healthMapper),
         HttpMethod.GET);
-    addRoute(
-        "/metrics", AdminHandlers.metricsHandler(metricRegistry, metricsMapper), HttpMethod.GET);
+    addRoute("/ping", AdminHandlers.pingHandler(), HttpMethod.GET);
+    addRoute("/ready", AdminHandlers.readyHandler(), HttpMethod.GET);
+    addRoute("/restart", AdminHandlers.restartHandler(this), HttpMethod.GET);
+    addRoute("/killkillkill", AdminHandlers.killHandler(this), HttpMethod.GET);
+
+    addRoute("/gc", AdminHandlers.pingHandler(), HttpMethod.GET);
```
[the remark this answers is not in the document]
xjdr: initiates a Garbage Collection sweep. I will add comments

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
xjdr: it is used to provide latency statistics for the individual reqs. As this is a shared handler, the ctx was needed as the key to ensure we got the correct timer out in a consistent and thread safe manner.

PR #68, opened 2018-01-03
build out tests for XUrl and decode query params

forplayplay:
This is a reboot of https://github.com/Nordstrom/xrpc/pull/65

- Adds more tests and resolves some exiting test issues.
- Decodes Query params and returns them in a data structure that fails gracefully in common use-cases. 
  
The new decodeQueryString method is slightly more performant that the previous, clocking on average 9ms vs 12ms

### DIFF
--- a/src/main/java/com/nordstrom/xrpc/client/DefaultValueMap.java
+++ b/src/main/java/com/nordstrom/xrpc/client/DefaultValueMap.java
@@ -3,7 +3,6 @@ package com.nordstrom.xrpc.client;
 import java.util.LinkedHashMap;
 
 /**
- *
  * A map that returns a default value instead of null when a given key is absent.
  *
  * This is the data structure returned for decoded query string parameters. It is meant to soften the edges around null-pointer errors.
--- a/src/main/java/com/nordstrom/xrpc/client/DefaultValueMap.java
+++ b/src/main/java/com/nordstrom/xrpc/client/DefaultValueMap.java
@@ -1,0 +1,22 @@
+package com.nordstrom.xrpc.client;
+
+import java.util.LinkedHashMap;
+
+/**
+ *
+ * A map that returns a default value instead of null when a given key is absent.
+ *
+ * This is the data structure returned for decoded query string parameters. It is meant to soften the edges around null-pointer errors.
+ */
+
+public class DefaultValueMap<K, V> extends LinkedHashMap<K, V> {
+  private final V defaultValue;
+  public DefaultValueMap(V defaultValue) {
+    this.defaultValue = defaultValue;
+  }
+
+  @Override
+  public V get(Object k) {
+    return containsKey(k) ? super.get(k) : defaultValue;
+  }
+}
--- a/src/main/java/com/nordstrom/xrpc/client/XUrl.java
+++ b/src/main/java/com/nordstrom/xrpc/client/XUrl.java
@@ -122,8 +122,7 @@ public class XUrl {
 
   public static Map<String, List<String>> decodeQueryString(String url) {
     Preconditions.checkNotNull(url);
-    // TODO(pdent): This should only operate on the path. Fix.
-    QueryStringDecoder decoder = new QueryStringDecoder(url);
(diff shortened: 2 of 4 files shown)


src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -118,13 +120,11 @@ public static String addProtocol(String url) {
...
-    return Arrays.stream(stripQueryParameters(url).split("&"))
-        .map(XUrl::splitQueryParameter)
-        .collect(
-            Collectors.groupingBy(
-                AbstractMap.SimpleImmutableEntry::getKey,
-                LinkedHashMap::new,
-                mapping(Map.Entry::getValue, toList())));
+    Preconditions.checkNotNull(url);
+    QueryStringDecoder decoder = new QueryStringDecoder(url);
+    Map<String, List<String>> params = new DefaultValueMap<>(new ArrayList<String>());
```
jkinkead: This should use `ImmutableList.of()` instead of an empty `ArrayList`.

src/main/java/com/nordstrom/xrpc/client/XUrl.java
```diff
@@ -118,13 +120,11 @@ public static String addProtocol(String url) {
...
   public static Map<String, List<String>> decodeQueryString(String url) {
-    return Arrays.stream(stripQueryParameters(url).split("&"))
-        .map(XUrl::splitQueryParameter)
-        .collect(
-            Collectors.groupingBy(
-                AbstractMap.SimpleImmutableEntry::getKey,
-                LinkedHashMap::new,
-                mapping(Map.Entry::getValue, toList())));
+    Preconditions.checkNotNull(url);
+    QueryStringDecoder decoder = new QueryStringDecoder(url);
```
jkinkead: This isn't meant to be used on a full URL, and might have bugs when processing a URL with a host+protocol. You should put a NOTE in noting that (`TODO(pdent): This should only operate on the path. Fix.`).

src/test/java/com/nordstrom/xrpc/client/XUrlTest.java
```diff
@@ -4,30 +4,42 @@
 
 import org.junit.jupiter.api.Test;
 
+import java.net.InetSocketAddress;
+
 class XUrlTest {
 
   String url1 = "https://api.nordstrom.com/foo/v1?foo=bar";
   String url2 = "https://api.nordstrom.com:8080/foo/v1?foo=bar";
+  String url3 = "https://api.nordstrom.com:8080/foo/bar%20baz?foo=bar";
```
jkinkead: Test style preference: I really don't like trivial fixtures like this, especially with unuseful names. Either inline `url3` below, or give it a descriptive name (`urlEncodedPath`).

Using a nondescriptive name just means that I have to jump back to the beginning of the class to interpret what a test is actually doing, and why it's failing.

It's one area where I feel strongly that good test style is the exact opposite of good regular code style (Do Repeat Yourself (In Tests)).

src/main/java/com/nordstrom/xrpc/client/DefaultValueMap.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.LinkedHashMap;
+
+/**
+ *
+ * A map that returns a default value instead of null when a given key is absent.
+ *
+ * This is the data structure returned for decoded query string parameters. It is meant to soften the edges around null-pointer errors.
+ */
+
```
jkinkead: Remove blank line?

src/test/java/com/nordstrom/xrpc/client/XrpcClientTest.java
```diff
@@ -9,6 +9,8 @@
 import java.net.URISyntaxException;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+
+import jdk.nashorn.internal.ir.annotations.Ignore;
```
jkinkead: Use [junit's `@Disabled`](http://junit.org/junit5/docs/current/user-guide/#writing-tests-annotations).

PR #69, opened 2018-01-03
Make Call class public to make methods usable.

spenserpothier:
* Also apply GoogleJavaFormat fix


src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -30,14 +30,15 @@
...
+
 import java.net.InetSocketAddress;
 import java.net.URISyntaxException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
-import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
-class Call {
+public class Call {
```
jkinkead: I don't think we want this class to be exposed as-is. The usage pattern is going to be confusing, and it's not thread-safe.

To fix, we probably want `request` to be `final`, and have `get` return a new instance of `Call`.

PR #72, opened 2018-01-12
Migrate releases to bintray.

jkinkead:
I also bumped the version to `0.1.5`, because that got missed during the last release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,54 +4,37 @@ For people contributing to this project.
 
 ## Publishing
 
-xrpc is published to [the central repository](https://oss.sonatype.org/#nexus-search;quick~com.nordstrom.xrpc).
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -4,60 +4,42 @@ For people contributing to this project.
 
 ## Publishing
 
-xrpc is published to [the central repository](https://oss.sonatype.org/#nexus-search;quick~com.nordstrom.xrpc).
+xrpc is published to [jCenter via Bintray](https://bintray.com/nordstromoss/oss_maven/xrpc).
```
andyday: does this mean we have a nordstrom bintray account now?

jkinkead: We have an open-source (free) organization.

I *think* this is a new feature, because I don't remember it being an option in the past.

PR #76, opened 2018-01-23
Modified XConfig to read cert and key from a path... 

lducharme:
...instead of from plaintext in config file.
However if no path is provided, XConfig falls back to default cert and key in xrpc.conf.


src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -127,6 +150,17 @@ private void populateClientOverrideList(List<? extends ConfigObject> req_per_sec
     return clientRateLimitOverride;
   }
 
+  private String readFromFile(Path path) {
+    String fileText = null;
+    try {
+      fileText = new String(Files.readAllBytes(path.toAbsolutePath()), XrpcConstants.DEFAULT_CHARSET);
+    } catch (IOException e) {
```
jkinkead: This should be considered a hard error, and should not be ignored. Either this should bubble out through the constructor, or it should be wrapped in a `RuntimeException` and rethrown (and documented).

demo/src/main/resources/demo.conf
```diff
@@ -8,11 +8,15 @@ xrpc {
   hard_req_per_sec = 550.0
   console_reporter = true
 
+  #path_to_cert = "/path/to/cert.txt"
```
jkinkead: You should add these to the main config file along with documentation.

lducharme: Added these configs and docs in the latest commit, thanks!

src/main/java/com/nordstrom/xrpc/XConfig.java
lducharme: Makes sense, I wrapped it and updated the docs.

demo/src/main/resources/demo.conf
```diff
@@ -15,4 +15,5 @@ xrpc {
   # IP White List
   enable_white_list = true
   ip_white_list = ["127.0.0.1"]
+
```
jkinkead: Revert newline here?

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -19,6 +19,10 @@
 import com.google.common.collect.ImmutableSet;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+import java.io.IOException;
```
jkinkead: Sort?

Running `./gradlew goJF` should do this for you.

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -87,10 +91,29 @@ public XConfig(Config configOverrides) {
     rateLimiterPoolSize = config.getInt("rate_limiter_pool_size");
     softReqPerSec = config.getDouble("soft_req_per_sec");
     hardReqPerSec = config.getDouble("hard_req_per_sec");
+
+    // Check to see if path_to_cert and path_to_key are configured; if so, use them. If not,
+    // fall back to cert and key configured in plaintext in xrpc.conf.
+    if (config.hasPath("path_to_cert")) {
+      // get and use non-null setting
```
jkinkead: Format comments as full sentences (initial capital, terminal period).

You also have a great block-level comment above, so you could probably skip the inline comments here.

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -127,6 +150,17 @@ private void populateClientOverrideList(List<? extends ConfigObject> req_per_sec
     return clientRateLimitOverride;
   }
 
+  private String readFromFile(Path path) {
+    String fileText;
+    try {
+      fileText = new String(Files.readAllBytes(path.toAbsolutePath()), XrpcConstants.DEFAULT_CHARSET);
+    } catch (IOException e) {
+      throw new RuntimeException("Could not read cert/key from path: " + path);
```
jkinkead: Use two-argument constructor to retain the cause of this exception (i.e. chain the exceptions).

src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -63,7 +63,18 @@ ip_black_list = []
 enable_white_list = false
 ip_white_list = []
 
+# Path to x509 certificate to use for TLS. Set to the absolute path of a Base64 encoded X509 certificate.
```
jkinkead: Key and cert are both in [PEM format](https://en.wikipedia.org/wiki/Privacy-enhanced_Electronic_Mail). You should probably just state that.

You might mention that the encoding is UTF-8, although there's essentially zero chance that will matter.

src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -63,7 +63,18 @@ ip_black_list = []
...
+# If this path and 'cert', below, are both configured, this path will be used and 'cert' will be ignored.
+# Throws a runtime error if a cert cannot be read from this path.
```
jkinkead: I forgot to mention:

This should go in the javadoc of the `XConfig` constructor instead. Something like:

```java
/**
 * @throws RuntimeException if there is an error reading one of path_to_cert or path_to_key
 */
```

lducharme:
Thanks for the comments! I think everything you mentioned should be fixed in the latest.

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -72,8 +79,10 @@ public XConfig() {
    * Construct a config object using the provided configuration, falling back on the default
    * configuration values <a
    * href="https://github.com/Nordstrom/xrpc/blob/master/src/main/resources/com/nordstrom/xrpc/xrpc.conf">here</a>.
+   *
+   * @throws RuntimeException if there is an error reading one of path_to_cert or path_to_key
    */
-  public XConfig(Config configOverrides) {
+  public XConfig(Config configOverrides) throws RuntimeException {
```
jkinkead: You don't need to add this to the declaration; it's usually considered sufficient to add the javadoc, like you did.

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -127,6 +151,17 @@ private void populateClientOverrideList(List<? extends ConfigObject> req_per_sec
     return clientRateLimitOverride;
   }
 
+  private String readFromFile(Path path) {
+    String fileText;
+    try {
+      fileText = new String(Files.readAllBytes(path.toAbsolutePath()), XrpcConstants.DEFAULT_CHARSET);
```
jkinkead: You could just return the string here directly, without the need for the local.

andyday:
:+1:  concur with jesse’s last comments

src/main/java/com/nordstrom/xrpc/XConfig.java
xjdr: Guava also has a nice utility for this.

```
import com.google.common.base.Charsets;
import com.google.common.io.Files;

String content = Files.toString(new File("/home/x1/text.log"), Charsets.UTF_8);
```

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -87,10 +96,25 @@ public XConfig(Config configOverrides) {
...
+      cert = config.getString("cert");
+    }
+
+    if (config.hasPath("path_to_key")) {
+      String pathToKey = config.getString("path_to_key");
+      key = readFromFile(Paths.get(pathToKey));
+    } else {
+      key = config.getString("key");
+    }
+
```
xjdr: You could also use a ternary operator here for a one liner:

```  
key = config.hasPath("path_to_cert") ?  Files.toString(new File(config.getString("path_to_cert")), Charsets.UTF_8) : config.getString("cert")```

PR #77, opened 2018-01-24
Add helpers to fetch request data and request headers.

jkinkead:
This is mostly in response to some silly boilerplate @forplayplay was needing to write.

This also decodes request bodies correctly, using any specified charset with the correct default, instead of always using UTF-8.


forplayplay:
LGTM!

andyday:
:+1:

PR #82, opened 2018-01-25
Additions for automated deployments to bintray via travis-ci

spenserpothier:


.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: java
+install: true
+script:
+  - ./gradlew clean check generatePomFileForXrpcMavenPublication publishToMavenLocal
```
jkinkead: I would make this `clean check test` (unless `check` subsumes `test`), and put the artifact-generation tasks in `before_deploy`.

andyday: check does subsume test.

andyday:
@jkinkead it is wise to avoid a specialized release branch if possible.  any shared branches add to the complexity of the lifecycle and merge process.  i would recommend doing a release on every successful merge to master.  this requires each PR to include CHANGELOG and set the new version if needed.  the version could also be auto incremented based on a Major, Minor, Patch indicator set by the PR creator...   this would also require automatic Github Release using API...   I wouldn't recommend this approach for larger projects like kubernetes or the like, but for a smaller library like this it makes sense to me at least.

jkinkead:
Right now, fully-automated builds aren't happening (it's been hard enough to get travis building). It might be a good next-step, but I'd like to keep the discussion in the context of this current change.

The issue here is that the current approach requires that, for a release to happen, the person running a release needs to have push permissions to `master` (needs to be a project admin). This isn't great. Having a shared release branch would fix this problem: Any person with normal write permissions could trigger a release.

It DOES add complexity, since it requires something like:
- configure `Nordstrom/xrpc` as `upstream`
- `git fetch upstream`
- `git branch -D release`
- `git checkout upstream/master`
- `git checkout -b release`
- `git push release upstream`
- `./gradlew release`

Instead of:
- configure `Nordstrom/xrpc` as `upstream`
- `git fetch upstream`
- `git checkout master`
- `./gradlew release`

The question is: Is the added complexity above worth the ability for non-maintainers to deploy?

.travis.yml
spenserpothier: moved the publish tasks to `before_deploy`

spenserpothier:
I think at this point the added complexity adds marginal value. I've added #83 to discuss the release branches, but personally I don't think this should be a blocker for this PR. As this PR currently stands, we at least get all future PRs built and tests run against them and just the version updates need to be done by a github `Nordstrom/xrpc` maintainer, but the rest is handled by travisci.
Also worth noting, Travis also includes a [github releases deployer](https://docs.travis-ci.com/user/deployment/releases/) that can be added.

jkinkead:
SGTM. Mostly I'd like for occasional contributors to be empowered to get their patches out without having to bother an admin.

andyday:
@jkinkead given the sequence you've listed above, if scripted and the fact that the release branch is a temporary branch it should do the trick...

PR #84, opened 2018-01-26
Add express style router sugar (aka. router.get, router.post, etc)

andyday:
Adding express-style router. get, post, put, etc.  The intent is to simplify usage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build/
 demo/out/
 out/
 xrpc.iml
+.DS_Store
--- a/demo/src/main/java/com/nordstrom/xrpc/demo/PeopleRoutes.java
+++ b/demo/src/main/java/com/nordstrom/xrpc/demo/PeopleRoutes.java
@@ -1,0 +1,61 @@
+package com.nordstrom.xrpc.demo;
+
+import com.nordstrom.xrpc.server.Router;
+import com.nordstrom.xrpc.server.http.Recipes;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.Moshi;
+import com.squareup.moshi.Types;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.AllArgsConstructor;
+
+public class PeopleRoutes {
+  final List<Person> people = new ArrayList<>();
+
+  public PeopleRoutes(Router router) {
+    // For the JSON portion of the demo
+    Moshi moshi = new Moshi.Builder().build();
+    Type type = Types.newParameterizedType(List.class, Person.class);
+    JsonAdapter<List<Person>> adapter = moshi.adapter(type);
+
+    router.get(
+        "/people/{person}",
+        request -> {
+          byte[] jsonBytes = adapter.toJson(people).getBytes(StandardCharsets.UTF_8);
+          return Recipes.newResponse(
+              HttpResponseStatus.OK,
+              request.getAlloc().directBuffer().writeBytes(jsonBytes),
+              Recipes.ContentType.Application_Json);
+        });
+
+    // Add handler for HTTP POST:/people route
+    router.post(
+        "/people",
+        request -> {
+          Person p = new Person(request.getDataAsString());
+          people.add(p);
+
+          return Recipes.newResponseOk("");
+        });
+
+    // Add handler for HTTP GET:/people route
+    router.get(
+        "/people",
+        request -> {
+          byte[] jsonBytes = adapter.toJson(people).getBytes(StandardCharsets.UTF_8);
+          return Recipes.newResponse(
+              HttpResponseStatus.OK,
+              request.getAlloc().directBuffer().writeBytes(jsonBytes),
+              Recipes.ContentType.Application_Json);
+        });
+  }
+
+  // Example POJO for use in request / response.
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ build/
 demo/out/
 out/
 xrpc.iml
+.DS_Store
```
jkinkead: This should go in your global gitignore.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -126,17 +131,50 @@ public void scheduleHealthChecks(
...
-            healthCheckRegistry.runHealthChecks(workerGroup);
-          }
-        },
-        initialDelay,
-        delay,
-        timeUnit);
+        () -> healthCheckRegistry.runHealthChecks(workerGroup), initialDelay, delay, timeUnit);
+  }
+
+  /** * Convenience Methods ** */
```
jkinkead: Shouldn't be javadoc-formatted.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -126,17 +131,50 @@ public void scheduleHealthChecks(
...
+  }
+
+  public void connect(String route, Handler handler) {
+    addRoute(route, handler, XHttpMethod.CONNECT);
+  }
+
+  public void any(String route, Handler handler) {
+    addRoute(route, handler, XHttpMethod.ANY);
+  }
+  /** ************************* */
```
jkinkead: Shouldn't be javadoc-formatted.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -126,17 +131,50 @@ public void scheduleHealthChecks(
...
-          }
-        },
-        initialDelay,
-        delay,
-        timeUnit);
+        () -> healthCheckRegistry.runHealthChecks(workerGroup), initialDelay, delay, timeUnit);
+  }
+
+  /** * Convenience Methods ** */
+  public void get(String route, Handler handler) {
```
jkinkead: It wouldn't hurt to add very simple javadoc to these, especially for ANY:

```java
/** Binds a default handler for all HTTP requests to the given route. */
...
/** Binds a handler for GET requests to the given route. */
```

.gitignore
andyday: i disagree.   project files should do as much as possible to make dev on-boarding simple.  the dev cost savings over time adds up...

src/main/java/com/nordstrom/xrpc/server/Router.java
andyday: duh... the google java formatter bastardized it...  will fix...

src/main/java/com/nordstrom/xrpc/server/Router.java
andyday: will do

src/main/java/com/nordstrom/xrpc/server/Router.java
andyday: see above...

forplayplay:
I think these helper methods are nice and simple!

I have mixed feelings about definition and registration all happening in main. It's great for small apps - but as they grow, I can see this pattern becoming cumbersome - especially for maintainers on-boarding further down the road. 

I would encourage us to separate handlers by resource in their own controller class. Is there some rational I am missing?

src/main/java/com/nordstrom/xrpc/server/Router.java
andyday: fixed

src/main/java/com/nordstrom/xrpc/server/Router.java
andyday: done

src/main/java/com/nordstrom/xrpc/server/Router.java
andyday: done

andyday:
i've updated based on review.   @forplayplay can you take a look at the revised example.  it shows a way to split routes based on resource group...

.gitignore
jkinkead: I'm not saying it shouldn't go here; I'm saying that you should have this in your global gitignore so you don't need to add it. :)

demo/src/main/java/com/nordstrom/xrpc/demo/PeopleRoutes.java
```diff
@@ -0,0 +1,61 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.AllArgsConstructor;
+
+public class PeopleRoutes {
+  final List<Person> people = new ArrayList<>();
+
+  public PeopleRoutes(Router router) {
+    // For the JSON portion of the demo
```
jkinkead: We should use Jackson; I'll open an issue to track this.

.gitignore
andyday: agreed.   added to global.  the one caveat to that is my adding it to global blinds me to the need for it in the project.  i've lead enough teams with junior devs that might mis ignoring such files/dirs and check them in.  having it in the project protects from such things...

demo/src/main/java/com/nordstrom/xrpc/demo/PeopleRoutes.java
andyday: yep.  will be doing this in my next pr.

PR #87, opened 2018-01-26
Added xio-0.13.0 as a dependency

pdex:


build.gradle
```diff
@@ -50,13 +51,15 @@ tasks.withType(JavaCompile) {
 
 repositories {
     mavenCentral()
+    maven { url 'https://dl.bintray.com/nordstromoss/oss_maven' }
```
andyday: using jcenter() as repo was sufficient for apim-api to get xrpc from bintray.  not sure why we would need any different here.

jkinkead: This project isn't synced to jcenter (yet). I assume @pdex has opened a ticket to get it syncing?

pdex: yes!

pdex: jcenter has approved my request, but xio hasn't landed there yet.

andyday:
excellent!  👍

PR #88, opened 2018-01-29
Break out server initializer

pdex:
The `Router` class has [a lot of functionality in it](https://en.wikipedia.org/wiki/God_object). This PR breaks out the embedded `ChannelInitializer` into the `ServerChannelInitializer` class.

### DIFF
--- a/src/main/java/com/nordstrom/xrpc/server/Router.java
+++ b/src/main/java/com/nordstrom/xrpc/server/Router.java
@@ -37,8 +37,6 @@ import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
-import io.netty.channel.ChannelInitializer;
-import io.netty.channel.ChannelPipeline;
 import io.netty.channel.EventLoopGroup;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
--- a/src/main/java/com/nordstrom/xrpc/server/Router.java
+++ b/src/main/java/com/nordstrom/xrpc/server/Router.java
@@ -30,7 +30,6 @@ import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Ordering;
 import com.nordstrom.xrpc.XConfig;
-import com.nordstrom.xrpc.logging.ExceptionLogger;
 import com.nordstrom.xrpc.server.http.Route;
 import com.nordstrom.xrpc.server.http.XHttpMethod;
 import com.nordstrom.xrpc.server.tls.Tls;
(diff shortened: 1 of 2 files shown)


src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public ServerChannelInitializer(
+      XConfig config,
+      ConnectionLimiter globalConnectionLimiter,
+      ServiceRateLimiter rateLimiter,
+      WhiteListFilter whiteListFilter,
+      BlackListFilter blackListFilter,
+      Firewall firewall,
+      Tls tls,
+      Http2OrHttpHandler h1h2) {
```
andyday: not a blocker but this number of constructor args is approaching too many

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
```diff
@@ -0,0 +1,82 @@
...
+    this.rateLimiter = rateLimiter;
+    this.whiteListFilter = whiteListFilter;
+    this.blackListFilter = blackListFilter;
+    this.firewall = firewall;
+    this.tls = tls;
+    this.h1h2 = h1h2;
+  }
+
+  @Override
+  public void initChannel(Channel ch) throws Exception {
```
andyday: since we are just creating a new class for this method, does it make sense to just make it a method on Router?  half of this class is boilerplate construction...

jkinkead: This looks like it should be a static method wherever it's placed.

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
```diff
@@ -0,0 +1,82 @@
...
+    cp.addLast(
+        "idleDisconnectHandler",
+        new IdleDisconnectHandler(
+            config.readerIdleTimeout(), config.writerIdleTimeout(), config.allIdleTimeout()));
+    cp.addLast("serverConnectionLimiter", globalConnectionLimiter);
+    cp.addLast("serverRateLimiter", rateLimiter);
+
+    if (config.enableWhiteList()) {
+      cp.addLast("whiteList", whiteListFilter);
+    } else if (config.enableBlackList()) {
```
jkinkead: This looks like a bug @xjdr ? I didn't catch this when it was added, but it's undocumented at the least.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -297,32 +294,15 @@ public void listenAndServe(boolean serveAdmin, boolean scheduleHealthChecks) thr
...
-            }
-
-            cp.addLast("firewall", firewall);
-            cp.addLast(
-                "encryptionHandler", tls.getEncryptionHandler(ch.alloc())); // Add Config for Certs
-            cp.addLast("codec", h1h2);
-            cp.addLast("exceptionLogger", new ExceptionLogger());
-          }
-        });
+        new ServerChannelInitializer(
```
jkinkead: I feel like virtually all of `listenAndServe` makes this class harder to read, but I'm not sure about a good split here.

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
andyday: that makes more sense...

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
andyday: what if instead of having Router initialize all of these and passing them to this class, having them initialized here?  Can pass config as arg...

andyday: or a decoupled xconfig

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
xjdr: Whitelist is block all except {[x]}, blacklist is block {[x]}, they are somewhat mutually exclusive and by doing the above you do not add handlers to the pipeline unnecessarily.

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
pdex: In `xio` we build a State object and pass it around. We can move towards that here if we need to.

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
jkinkead: I see. This whole time I thought that whitelist meant that you were carving exceptions out from a blacklist...

andyday:
👍 non of my comments are blockers

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
pdex: @jkinkead `initChannel` is defined in netty's `ChannelInitializer`. I'm a little confused as to what you want to make into a `static` method

src/main/java/com/nordstrom/xrpc/server/Router.java
pdex: see [my other comment](https://github.com/Nordstrom/xrpc/pull/88#discussion_r164529770) I want to be able to customize the pipeline.

src/main/java/com/nordstrom/xrpc/server/ServerChannelInitializer.java
jkinkead: @pdex I meant that this logic could be adapted into a static method - but yes, it makes sense that you'd need to override the init method for the actual initializer.

You can disregard my comment; I wrote it before I saw your reply.

PR #89, opened 2018-01-29
Add and fix compiler warnings.

jkinkead:
There were a number of unchecked operations (especially on `RendezvousHash` usage), and a few calls to deprecated methods.

`addCallback` in particular is slated to be removed in April.

### DIFF
--- a/src/main/java/com/nordstrom/xrpc/client/Call.java
+++ b/src/main/java/com/nordstrom/xrpc/client/Call.java
@@ -20,6 +20,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.SettableFuture;
 import com.nordstrom.xrpc.client.retry.BoundedExponentialBackoffRetry;
 import com.nordstrom.xrpc.client.retry.RetryLoop;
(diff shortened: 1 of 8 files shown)


src/main/java/com/nordstrom/xrpc/client/Call.java
```diff
@@ -91,7 +92,9 @@ public void onFailure(Throwable t) {
             response.cancel(true);
             error.setException(t);
           }
-        });
+        },
+        MoreExecutors.directExecutor()
```
andyday: can you explain this usage?  not that obvious...

src/test/java/com/nordstrom/xrpc/client/XrpcClientTest.java
```diff
@@ -63,7 +64,9 @@ public void onFailure(Throwable t) {
             latch.countDown();
             assertEquals(true, false);
           }
-        });
+        },
+        MoreExecutors.directExecutor()
```
andyday: see comment above...

src/main/java/com/nordstrom/xrpc/client/Call.java
jkinkead: [See the deprecation note](http://google.github.io/guava/releases/snapshot-jre/api/docs/com/google/common/util/concurrent/Futures.html#addCallback-com.google.common.util.concurrent.ListenableFuture-com.google.common.util.concurrent.FutureCallback-). This is replacing what was an implicit executor with the explicit version.

The concept of a direct (threadless) executor is used all over when operating within the java.util.concurrent world.

andyday:
formatting error -> travis

PR #91, opened 2018-01-29
Created server State object

pdex:


src/main/java/com/nordstrom/xrpc/server/State.java
```diff
@@ -0,0 +1,66 @@
...
+
+package com.nordstrom.xrpc.server;
+
+import com.google.auto.value.AutoValue;
+import com.nordstrom.xrpc.XConfig;
+import com.nordstrom.xrpc.server.tls.Tls;
+
+/** State is a value object for storing server state */
+@AutoValue
+public abstract class State {
```
andyday: Is it intended that this could change dynamically after config and startup?  If not, could this be merged with XConfig?

src/main/java/com/nordstrom/xrpc/server/State.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright 2017 Nordstrom, Inc.
```
andyday: we should probably update this thru 2018

src/main/java/com/nordstrom/xrpc/server/State.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public abstract ConnectionLimiter globalConnectionLimiter();
+
+  public abstract ServiceRateLimiter rateLimiter();
+
+  public abstract WhiteListFilter whiteListFilter();
+
+  public abstract BlackListFilter blackListFilter();
+
+  public abstract Firewall firewall();
```
andyday: in my view firewall should include whiteList and blackList...

src/main/java/com/nordstrom/xrpc/server/State.java
pdex: @jkinkead can we use something like [this](https://github.com/hierynomus/license-gradle-plugin) to manage copyright years?

src/main/java/com/nordstrom/xrpc/server/State.java
pdex: This is a value object that represents server state, in general I try to keep config data separate from running state.

src/main/java/com/nordstrom/xrpc/server/State.java
pdex: Should we track that in an issue?

src/main/java/com/nordstrom/xrpc/server/State.java
andyday: +1 for this

src/main/java/com/nordstrom/xrpc/server/State.java
andyday: probably so...

src/main/java/com/nordstrom/xrpc/server/State.java
andyday: if it is running state that can be changed after startup, should we use atomic refs here?

src/main/java/com/nordstrom/xrpc/server/State.java
xjdr: firewall (as of now) is only managing the blocking. The different filters create the events to inform the firewall when to block

src/main/java/com/nordstrom/xrpc/server/State.java
xjdr: The server `State` should be static and as of now the XConfig only loads once on startup. We can create an issue to have multiple providers of dynamic config info and wrap that dynamic config in a AtomicRef

PR #95, opened 2018-01-30
Add HttpQuery and XrpcRequest methods to help with query and responses

andyday:


src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -16,24 +16,36 @@
...
 import java.nio.charset.Charset;
+import java.util.Collections;
 import java.util.Map;
 import lombok.Getter;
+import lombok.experimental.Accessors;
+import lombok.val;
 
 /** Xprc specific Request object. */
 public class XrpcRequest {
+  private static final ObjectMapper mapper = new ObjectMapper();
```
jkinkead: This should be injected into the constructor. We want to be able to support custom mappers, and this precludes that.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,62 @@
...
+  public String rawPath() {
+    return decoder().rawPath();
+  }
+
+  /** Get raw query string */
+  public String rawQuery() {
+    return decoder().rawQuery();
+  }
+
+  private QueryStringDecoder decoder() {
```
jkinkead: This isn't thread-safe. That's fine, but should be noted in class-level Javadoc.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,62 @@
+package com.nordstrom.xrpc.server;
+
+import io.netty.handler.codec.http.QueryStringDecoder;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import lombok.val;
+
+/** Parsed query string */
```
jkinkead: Javadoc should end in terminal period, here and elsewhere in this class.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,62 @@
...
+public class HttpQuery {
+  private String uri;
```
jkinkead: `final`?

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -44,6 +56,11 @@
   /** The variables captured from the route path. */
   private final Map<String, String> groups;
 
+  /** The parsed query string. */
+  @Getter
+  @Accessors(fluent = true)
+  private final HttpQuery query;
```
jkinkead: I think I'd prefer that this be lazily-initialized, and have the `HttpQuery` class be a simple wrapper around `QueryStringDecoder`. Thoughts?

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,62 @@
...
+  /** Get a query parameter by key */
+  public Optional<String> parameter(String key) {
+    return parameter(key, null);
+  }
+
+  /** Get a query parameter by key with a default value */
+  public Optional<String> parameter(String key, String defaultValue) {
+    val value = parameters().get(key);
+    if (value == null || value.size() < 1) return Optional.of(defaultValue);
+    return Optional.of(value.get(0));
```
jkinkead:
```java
return Optional.of(parameters.get(key)).flatMap(list -> list.stream().findFirst());
```

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -66,6 +84,7 @@ public XrpcRequest(
     this.alloc = channel.alloc();
     this.eventLoop = channel.eventLoop();
     this.streamId = streamId;
+    this.query = new HttpQuery("");
```
jkinkead: This won't work for HTTP/2. This is a pretty good argument for making this be a lazily-initialized field...

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -153,4 +172,40 @@ public FullHttpRequest getHttpRequest() {
...
+  public FullHttpResponse notFoundJsonResponse(Object body) {
+    return jsonResponse(HttpResponseStatus.NOT_FOUND, body);
+  }
+
+  public FullHttpResponse badRequestJsonResponse(Object body) {
+    return jsonResponse(HttpResponseStatus.BAD_REQUEST, body);
+  }
+
+  private ByteBuf bodyToByteBuf(Object body) {
+    val buf = getAlloc().directBuffer();
```
jkinkead: We haven't been using `val` in xrpc. It might be better to leave it out of this PR and add it in later, everywhere.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -153,4 +172,40 @@ public FullHttpRequest getHttpRequest() {
...
+    OutputStream stream = new ByteBufOutputStream(buf);
+
+    try {
+      mapper.writeValue(stream, body);
+      return buf;
+    } catch (IOException e) {
```
jkinkead: I think it would be better to bubble this out, since the `Handler` interface allows `IOException` to be thrown.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
jkinkead: Actually, I forgot the default.
```java
return Optional.of(parameters.get(key)).flatMap(list -> list.stream().findFirst()).orElse(defaultValue);
```
That returns `String`, not `Optional<String>`, but that seems correct for this method?

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: done

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: done

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: nice catch. done.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: this has been removed.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: done.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: done. this is now lazy initialized and throws exception if used with http2

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: removed

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: done.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: done.

andyday:
@jkinkead all comments have been addressed.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /** Get a query parameter by key. */
+  public Optional<String> parameter(String key) {
+    return parameter(key, null);
+  }
+
+  /** Get a query parameter by key with a default value. */
+  public Optional<String> parameter(String key, String defaultValue) {
+    return Optional.of(
+        Optional.of(parameters().get(key))
```
jkinkead: My error: This should be `ofNullable`, not `of`.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,54 @@
...
+    return decoder.parameters();
+  }
+
+  /** Get a query parameter by key. */
+  public Optional<String> parameter(String key) {
+    return parameter(key, null);
+  }
+
+  /** Get a query parameter by key with a default value. */
+  public Optional<String> parameter(String key, String defaultValue) {
```
jkinkead: It's weird that this returns `Optional`, given that you supply a default.

I'd have this return a `String`, and have `parameter` return `Optional`. Then you'd have:

```java
public Optional<String> parameter(String key) {
  return Optional.ofNullable(parameters().get(key)).flatMap(list -> list.stream().findFirst());
}

public String parameter(String key, String defaultValue) {
  return parameter(key).orElse(defaultValue);
}
```

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,54 @@
...
+  public String path() {
+    return decoder.path();
+  }
+
+  /** Map of query parameters. */
+  public Map<String, List<String>> parameters() {
+    return decoder.parameters();
+  }
+
+  /** Get a query parameter by key. */
```
jkinkead: Document that this returns the first value supplied in a URL, and ignores any extra values provided?

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -57,16 +72,31 @@ public XrpcRequest(FullHttpRequest request, Map<String, String> groups, Channel
...
     this.groups = groups;
     this.upstreamChannel = channel;
     this.alloc = channel.alloc();
     this.eventLoop = channel.eventLoop();
     this.streamId = streamId;
   }
 
+  public HttpQuery query() {
+    if (h1Request == null) {
+      throw new UnsupportedOperationException("query() is not supported for HTTP2.");
```
jkinkead: This should pull the query out of `h2Headers.path()`:
```java
query = new HttpQuery(h2Request.path());
```

andyday: duh.  don't know what I was thinking.  done.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: yuck.  good catch...  done.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: agreed.  fixed.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: done.

andyday:
fixed latest review comments.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /** Map of query parameters. */
+  public Map<String, List<String>> parameters() {
+    return decoder.parameters();
+  }
+
+  /** Get a the first query parameter by key, ignoring any extras. */
+  public Optional<String> parameter(String key) {
+    return Optional.ofNullable(parameter(key, null));
```
jkinkead: This should be `parameters.get(key)`, not `parameter(key, null)`; as-is will throw a `NullPointerException`.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -57,16 +72,32 @@ public XrpcRequest(FullHttpRequest request, Map<String, String> groups, Channel
...
     this.streamId = streamId;
   }
 
+  public HttpQuery query() {
+    if (query == null) {
+      if (h1Request != null) {
+        query = new HttpQuery(h1Request.uri());
+      } else if (h2Headers != null) {
+        query = new HttpQuery(h2Headers.path().toString());
+      }
```
jkinkead: `else` throw an `IllegalStateException`? That's how we usually handle this.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -57,16 +72,32 @@ public XrpcRequest(FullHttpRequest request, Map<String, String> groups, Channel
...
     this.eventLoop = channel.eventLoop();
     this.streamId = streamId;
   }
 
+  public HttpQuery query() {
+    if (query == null) {
+      if (h1Request != null) {
+        query = new HttpQuery(h1Request.uri());
+      } else if (h2Headers != null) {
+        query = new HttpQuery(h2Headers.path().toString());
```
jkinkead: Instead update `HttpQuery` to take a `CharSequence` argument?

Netty has a lot of optimizations around not using `java.lang.String`, and accommodating them helps speed.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: tested it and it doesn't throw an exception. works as expected

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: in this case the underlying decoder expects a String.  i'd have to upcast or convert as some point anyway.  best to do it at the point of entry.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: done

andyday:
@jkinkead added unit tests for HttpQuery and fixed last comment.  i've replied to the other comments.

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
jkinkead: That's a broken `orElse` API. :)

jkinkead: I would still change this, since the point of `Optional` is to avoid `null`, but it's not a big deal.

src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
jkinkead: SGTM!

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
andyday: per the javadocs:

```java
    /**
     * Return the value if present, otherwise return {@code other}.
     *
     * @param other the value to be returned if there is no value present, may
     * be null
     * @return the value, if present, otherwise {@code other}
     */
```

i tested that if the value does not exist that it returns Optional.empty().  based on my understanding of Optional, this is perfectly expectable.  same is true in kotlin and c# ...

andyday:
@jkinkead do you approve your prior review?  shows you approved but one review still outstanding...

andyday:
nvm

src/main/java/com/nordstrom/xrpc/server/HttpQuery.java
jkinkead: What's broken (to me) is that `orElse` is within the domain of `Optional`, and should have similar semantics. You can't have a present-but-`null` `Optional`, and it's deeply confusing that the unwrapping allows for an illegal value.

In my mind, `orElse` should behave identically to:
```java
Optional<T> foo;
Optional<T> bar;

if (foo.isPresent()) {
  return foo.get();
} else {
  bar.get();
}
```
Anyway, it's not the only language that does it this way, I just don't usually see it used.

PR #96, opened 2018-01-30
Custom pipeline demo

pdex:


demos/request-response-pipeline/build.gradle
```diff
@@ -0,0 +1,68 @@
+// Load plugins.
+buildscript {
```
andyday: Now that we have more than one demo, can we move the original demo under this new demos folder?

demos/request-response-pipeline/src/main/java/com/nordstrom/xrpc/pipelinedemo/Example.java
```diff
@@ -0,0 +1,135 @@
...
+    private final State state;
+    private final ImmutableMap<Route, PipelineRequestHandler> routes;
+
+    public MyChannelInitializer(State state, ImmutableMap<Route, PipelineRequestHandler> routes) {
+      this.state = state;
+      this.routes = routes;
+    }
+
+    @Override
+    public void initChannel(Channel ch) throws Exception {
```
andyday: since demos are for the average (non-netty-ified) dev, it might be good to add some comments as to the purpose and usage here...

pdex: The intention here is to get people interacting with the xio Request/Response objects. We can move that ChannelInitializer somewhere else.

andyday: that's fine.  should add some comments to that effect...

andyday:
👍 - my comments are non-blocking...

PR #101, opened 2018-01-31
Add CORS Support

andyday:


src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -50,7 +58,7 @@
   private final String cert;
   private final String key;
   private final int port;
-  private final double gloablSoftReqPerSec;
+  private final double globalSoftReqPerSec;
```
jkinkead: Hah! Good catch. :)

src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -63,6 +63,14 @@ ip_black_list = []
 enable_white_list = false
 ip_white_list = []
 
+# CORS
+enable_cors = false
+cors_allowed_origins = []
+cors_allowed_headers = []
```
jkinkead: The correct default for headers & methods is usually `*` (all). Does setting an empty array return this from the server?

src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -63,6 +63,14 @@ ip_black_list = []
 enable_white_list = false
 ip_white_list = []
 
+# CORS
+enable_cors = false
```
jkinkead: I really like using the nested config structure typesafe config supports:
```
cors = {
  enable = false
  allowed_origins = []
  ...
}
```
You then read the keys `cors.enable`, `cors.allowed_origins`, etc.

This makes things more organized and readable - and you can do cool things like pulling out the whole `cors` block as a `Config` object and pass it to a helper.

src/main/java/com/nordstrom/xrpc/server/Http2OrHttpHandler.java
```diff
@@ -52,6 +57,10 @@ protected void configurePipeline(ChannelHandlerContext ctx, String protocol) thr
       ChannelPipeline cp = ctx.pipeline();
       cp.addLast("codec", new HttpServerCodec());
       cp.addLast("aggregator", new HttpObjectAggregator(MAX_PAYLOAD_SIZE));
+
+      if (corsConfig.isCorsSupportEnabled()) {
```
jkinkead: Any reason not to return CORS headers always, with restrictive defaults? Or at least default to on?

src/test/java/com/nordstrom/xrpc/server/OkHttpUnsafe.java
```diff
@@ -0,0 +1,113 @@
...
+      throws NoSuchAlgorithmException, KeyManagementException {
+    // Create a trust manager that does not validate certificate chains
+    final TrustManager[] trustAllCerts = new TrustManager[] {trustManager};
+
+    // Install the all-trusting trust manager
+    final SSLContext sslContext = SSLContext.getInstance("TLSv1.2");
+    sslContext.init(keyManagers, trustAllCerts, new java.security.SecureRandom());
+    // Create an ssl socket factory with our all-trusting manager
+    final SSLSocketFactory sslSocketFactory = sslContext.getSocketFactory();
+    // System.out.println("supported ciphers: " +
```
jkinkead: Remove commented-out code?

src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -63,6 +63,14 @@ ip_black_list = []
...
+cors_allowed_origins = []
+cors_allowed_headers = []
+cors_allowed_methods = []
+cors_allow_credentials = false
```
jkinkead: I would document this and `short_circuit` below, since they are non-obvious.

It's not clear to me how `short_circuit` would affect interaction with a browser, or what exactly it does. I think it only affects `OPTIONS` requests, and might actually be a nice safeguard against accidental root ANY handlers.

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -37,6 +43,8 @@
  * href="https://github.com/Nordstrom/xrpc/blob/master/src/main/resources/com/nordstrom/xrpc/xrpc.conf">the
  * embedded config</a> for defaults and documentation.
  */
+@Accessors(fluent = true)
```
jkinkead: Thumbs-up emoji.

src/main/java/com/nordstrom/xrpc/server/Http2OrHttpHandler.java
andyday: probably not.  most frameworks make cors an opt-in experience.  for those that don't need it its just one more bit of overhead.

src/main/resources/com/nordstrom/xrpc/xrpc.conf
andyday: agreed was just trying to follow the current conventions..  will change.

src/main/resources/com/nordstrom/xrpc/xrpc.conf
andyday: i don't think we ever want to return `*`.  if the methods list is empty it will not return the allow methods header which will use the browser defaults...

src/test/java/com/nordstrom/xrpc/server/OkHttpUnsafe.java
andyday: ha. copied this from nfe...  will do...

src/main/resources/com/nordstrom/xrpc/xrpc.conf
andyday: allow_credentials is a standard cors header, but i'll add documentation for both.  i was toying with the idea of defaulting short circuit to true..

src/test/java/com/nordstrom/xrpc/server/OkHttpUnsafe.java
andyday: fixed

src/main/java/com/nordstrom/xrpc/server/Http2OrHttpHandler.java
andyday: for now leaving this as opt-in

andyday:
have made fixes based on review...

bin/corstest.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash -e
+
+LINE_BREAK="======================================"
```
jkinkead: It's not entirely clear why this file is in the repo; document?

demo/src/main/resources/demo.conf
```diff
@@ -15,4 +15,7 @@ xrpc {
   # IP White List
   enable_white_list = true
   ip_white_list = ["127.0.0.1"]
+
+  enable_cors = true
```
jkinkead: Update the demo app config?

src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -129,9 +139,41 @@ public XConfig(Config configOverrides) {
     ipWhiteList =
         ImmutableSet.<String>builder().addAll(config.getStringList("ip_white_list")).build();
 
+    corsConfig = configureCors(config.getConfig("cors"));
+
     populateClientOverrideList(config.getObjectList("req_per_second_override"));
   }
 
+  private CorsConfig configureCors(Config config) {
```
jkinkead: Style suggestion: Name `getCorsConfig` or `buildCorsConfig`.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -95,19 +97,19 @@ private void configResponseCodeMeters() {
...
-    final String NAME_PREFIX = "responseCodes.";
-    meterNamesByStatusCode.put(HttpResponseStatus.OK, NAME_PREFIX + "ok");
-    meterNamesByStatusCode.put(HttpResponseStatus.CREATED, NAME_PREFIX + "created");
-    meterNamesByStatusCode.put(HttpResponseStatus.ACCEPTED, NAME_PREFIX + "accepted");
-    meterNamesByStatusCode.put(HttpResponseStatus.NO_CONTENT, NAME_PREFIX + "noContent");
-    meterNamesByStatusCode.put(HttpResponseStatus.BAD_REQUEST, NAME_PREFIX + "badRequest");
-    meterNamesByStatusCode.put(HttpResponseStatus.UNAUTHORIZED, NAME_PREFIX + "unauthorized");
-    meterNamesByStatusCode.put(HttpResponseStatus.FORBIDDEN, NAME_PREFIX + "forbidden");
-    meterNamesByStatusCode.put(HttpResponseStatus.NOT_FOUND, NAME_PREFIX + "notFound");
+    final String namePrefix = "responseCodes.";
```
jkinkead: If you wanted to keep the name, this could be a `static final` member.

andyday: left this as final to keep it closer to its usage

andyday:
fixed latest review comments

Issue #103, opened 2018-02-02
Metrics namespacing inconsistencies


spenserpothier:
### Expected behavior
Consistent naming between metrics

### Actual behavior
Inconsistent naming between metrics

### Steps to reproduce
`/metrics` endpoint

### Minimal yet complete reproducer code (or URL to code)
https://github.com/Nordstrom/xrpc/pull/102#discussion_r165738853

### xrpc version
0.2.1

Sample output of metrics endpoint
---

```json
{
  "version" : "3.1.3",
  "gauges" : { },
  "counters" : {
    "com.nordstrom.xrpc.server.Router.Active Connections" : {
      "count" : 1
    }
  },
  "histograms" : { },
  "meters" : {
    "Hard Rate Limits" : {
      "count" : 0,
      "m15_rate" : 0.0,
      "m1_rate" : 0.0,
      "m5_rate" : 0.0,
      "mean_rate" : 0.0,
      "units" : "events/second"
    },
    "com.nordstrom.xrpc.server.Router.requests.Rate" : { 
      "count" : 2,
      "m15_rate" : 4.405910322420641E-4,
      "m1_rate" : 1.5708635764593015E-8,
      "m5_rate" : 2.0783399443213604E-4,
      "mean_rate" : 0.0023767358720068054,
      "units" : "events/second"
    },
    "requests" : {
      "count" : 0,
      "m15_rate" : 0.0,
      "m1_rate" : 0.0,
      "m5_rate" : 0.0,
      "mean_rate" : 0.0,
      "units" : "events/second"
    },
    "responseCodes.accepted" : {
      "count" : 0,
      "m15_rate" : 0.0,
      "m1_rate" : 0.0,
      "m5_rate" : 0.0,
      "mean_rate" : 0.0,
      "units" : "events/second"
    },
    "responseCodes.badRequest" : {
      "count" : 0,
      "m15_rate" : 0.0,
      "m1_rate" : 0.0,
      "m5_rate" : 0.0,
      "mean_rate" : 0.0,
      "units" : "events/second"
    },
    "responseCodes.created" : {
      "count" : 0,
      "m15_rate" : 0.0,
      "m1_rate" : 0.0,
      "m5_rate" : 0.0,
      "mean_rate" : 0.0,
      "units" : "events/second"
    },
    "responseCodes.forbidden" : {
      "count" : 0,
      "m15_rate" : 0.0,
      "m1_rate" : 0.0,
      "m5_rate" : 0.0,
[...]

PR #104, opened 2018-02-02
Added endpoint-level metrics keeping track of the request rate for each path.

lducharme:
I added metrics for request rate at the endpoint level. In addition to having this:
```
com.nordstrom.xrpc.server.Router.requests.Rate
             count = 0
         mean rate = 0.00 events/second
     1-minute rate = 0.00 events/second
     5-minute rate = 0.00 events/second
    15-minute rate = 0.00 events/second
```

we now also have metrics that keep track of the rate of requests for each endpoint (that is, combination of HTTP method and path):

```
com.nordstrom.xrpc.server.Router.handler.rate.get.people
             count = 0
         mean rate = 0.00 events/second
     1-minute rate = 0.00 events/second
     5-minute rate = 0.00 events/second
    15-minute rate = 0.00 events/second
com.nordstrom.xrpc.server.Router.handler.rate.post.people.{person}
             count = 0
         mean rate = 0.00 events/second
     1-minute rate = 0.00 events/second
     5-minute rate = 0.00 events/second
    15-minute rate = 0.00 events/second

```

A few things I'm unsure about and would especially appreciate feedback on:
Is ServiceRateLimiter the right place to configure these metrics? Since the overall server level rate is configured there, that's what I went with, but it seems a little off.

Re: [issue 103](https://github.com/Nordstrom/xrpc/issues/103), these metric names might be contributing to further inconsistency- any thoughts on naming these better?

Note we also may consider adding metrics per client. If we have metrics per client and per endpoint, this could end up being a large/unreadable number of metrics... how might we deal with this?


src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class MetricsUtil {
+
+  protected static String getMeterNameForRoute(Route route, XHttpMethod httpMethod) {
+    return getMeterNameForRoute(route, httpMethod.name());
+  }
+
+  protected static String getMeterNameForRoute(Route route, String httpMethod) {
+    String result = "";
+    if (httpMethod != null) {
```
andyday: we should default this to 'ANY'

src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
```diff
@@ -0,0 +1,24 @@
...
+      result += httpMethod.toLowerCase();
+    }
+
+    if (route != null && route.toString() != null) {
```
andyday: no route should cause us to throw a new ArgumentNullException

src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
```diff
@@ -0,0 +1,24 @@
...
+import com.nordstrom.xrpc.server.http.Route;
+import com.nordstrom.xrpc.server.http.XHttpMethod;
+
+public class MetricsUtil {
+
+  protected static String getMeterNameForRoute(Route route, XHttpMethod httpMethod) {
+    return getMeterNameForRoute(route, httpMethod.name());
+  }
+
+  protected static String getMeterNameForRoute(Route route, String httpMethod) {
```
andyday: since httpMethod can be null we should use `Optional<String> httpMethod` here

src/main/java/com/nordstrom/xrpc/server/XrpcConnectionContext.java
```diff
@@ -46,4 +46,8 @@
       routes = new AtomicReference<>();
 
   @Getter private final ObjectMapper mapper;
+
+  @Getter
+  private final ConcurrentHashMap<String, Meter> rateMetersByRouteAndMethod =
```
andyday: this should be named something like `metersByRoute` .  rate doesn't apply and Method is unneeded.  ideally, method is already part of the route (conceptually)

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -63,6 +69,32 @@ public ServiceRateLimiter(MetricRegistry metrics, XConfig config) {
         buildHasher(softLimiterMap, config.getRateLimiterPoolSize(), config.softReqPerSec());
     hardRateLimitHasher =
         buildHasher(hardLimiterMap, config.getRateLimiterPoolSize(), config.hardReqPerSec());
+    configEndpointLevelRateMeters(metrics, ctx);
+  }
+
+  private void configEndpointLevelRateMeters(
```
andyday: this doesn't belong in this class.  it should be done just below the `configResponseCodeMeters()` in Router

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -63,6 +69,32 @@ public ServiceRateLimiter(MetricRegistry metrics, XConfig config) {
...
+      MetricRegistry metricRegistry, XrpcConnectionContext ctx) {
+    ImmutableSortedMap<Route, List<ImmutableMap<XHttpMethod, Handler>>> routes =
+        ctx.getRoutes().get();
+
+    final String NAME_PREFIX = "handler.rate.";
```
andyday: this should be changed to something like `"routes.";`  also the name should be changed to `namePrefix`

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -63,6 +69,32 @@ public ServiceRateLimiter(MetricRegistry metrics, XConfig config) {
...
+
+    Set<String> routeAndMethodNames = new HashSet<>();
```
andyday: no need to have this temp holding set.  the routes can be added directly to the metric registry

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
```diff
@@ -63,6 +69,32 @@ public ServiceRateLimiter(MetricRegistry metrics, XConfig config) {
...
+      for (ImmutableMap<XHttpMethod, Handler> map : entry.getValue()) {
+        for (XHttpMethod httpMethod : map.keySet()) {
+          routeAndMethodNames.add(MetricsUtil.getMeterNameForRoute(route, httpMethod));
+        }
+      }
+    }
+
+    for (String routeName : routeAndMethodNames) {
+      ctx.getRateMetersByRouteAndMethod()
+          .put(routeName, metricRegistry.meter(name(Router.class, NAME_PREFIX + routeName)));
```
andyday: change to `.meter(namePrefix + routeName)...` .   this includes my other comment about changing `NAME_PREFIX` to `namePrefix`

src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java
lducharme: I originally had it where you suggest, in the constructor for Router, but at that point the routes aren't configured yet, so I can't set up meters for each route.

Looking at it again, perhaps a more appropriate place would be to put it would be early in Router.listenAndServe?

xjdr:
I'd like to see the regressions numbers between this and Master to make sure we aren't adding additional latency with this.

lducharme:
Thanks for the comments, Andy. I moved the metrics config out of ServiceRateLimiter and into Router.listenAndServe, which is hopefully a better place for it. I also updated the variable names and metric names, and changed MetricsUtil to use Optional to handle possible null parameters.

Let me know if any of this still has issues. I am working on doing benchmarking to check on the latency.

src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
```diff
@@ -10,15 +11,15 @@ protected static String getMeterNameForRoute(Route route, XHttpMethod httpMethod
...
+    String routeIdentifier;
     if (route != null && route.toString() != null) {
-      result += route.toString().replace('/', '.');
+      routeIdentifier = route.toString().replace('/', '.');
+    } else {
+      throw new IllegalArgumentException("Route cannot be null.");
     }
 
-    return result;
+    return method + routeIdentifier;
```
andyday: this return can happen inside the if, eliminating the need for the routeIdentifier variable.  in addition it is a good practice to use inverted if conditions to allow the happy path logic to happen outside of nested conditionals.  for instance...

```java
if (route == null || route.toString() == null) {
  throw new IllegalArgumentException("Route cannot be null.");
}

return String.format("{}{}", method, route.toString().replace('/', '.');
```

is better than the alternative you've presented.  it allows the happy path to be as unencumbered by levels of nesting while reducing the overall Cyclomatic Complexity of the method.

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -365,6 +369,26 @@ public void listenAndServe(boolean serveAdmin, boolean scheduleHealthChecks) thr
     channel = future.channel();
   }
 
+  private void configEndpointLevelRateMeters(
```
andyday: I'm not sure that these are accurately Rate meters in the same sense as we use the word Rate in `Rate Limiting`.    Or at minimum they are using different Rates.   That being said we should use a different word here to avoid confusion.  Some suggestions `ThroughPut`, `RequestCount`, `Requests`....

src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -365,6 +369,26 @@ public void listenAndServe(boolean serveAdmin, boolean scheduleHealthChecks) thr
...
+    final String namePrefix = "routes.";
+
+    for (Map.Entry<Route, List<ImmutableMap<XHttpMethod, Handler>>> entry : routes.entrySet()) {
+      Route route = entry.getKey();
+
+      for (ImmutableMap<XHttpMethod, Handler> map : entry.getValue()) {
+        for (XHttpMethod httpMethod : map.keySet()) {
+          String routeName = MetricsUtil.getMeterNameForRoute(route, httpMethod);
+          ctx.getMetersByRoute()
+              .put(routeName, metricRegistry.meter(name(Router.class, namePrefix + routeName)));
```
andyday: this should use the .meter method that uses only `namePrefix + routeName` as the name without Router.class

andyday: using the class makes the name overly verbose and noisy

lducharme: Sounds good, I'll remove it.

I originally included the class name because ServiceRateLimiter did it, here: 
https://github.com/Nordstrom/xrpc/blob/master/src/main/java/com/nordstrom/xrpc/server/ServiceRateLimiter.java#L56
Does it also make sense to remove the class name there, too?

lducharme:
I used the run regression script to check out latency in master vs my branch. The max is definitely higher but the average is very close. (Note I had to increase the server and global rate limits for this to work):

master:
```
  4 threads and 15 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.44ms    2.89ms  62.62ms   92.39%
    Req/Sec   312.18    208.79     1.32k    80.91%
  Latency Distribution
     50%    1.70ms
     75%    3.37ms
     90%    4.97ms
     99%    9.21ms
  54451 requests in 1.00m, 3.74MB read
  Socket errors: connect 12, read 54451, write 0, timeout 0
Requests/sec:    905.97
Transfer/sec:     63.70KB
```

my branch:
```
  4 threads and 15 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.46ms    3.18ms  87.48ms   93.80%
    Req/Sec   312.49    200.63     1.18k    79.12%
  Latency Distribution
     50%    1.68ms
     75%    3.30ms
     90%    4.93ms
     99%    9.37ms
  56996 requests in 1.00m, 3.91MB read
  Socket errors: connect 12, read 56996, write 0, timeout 0
Requests/sec:    948.68
Transfer/sec:     66.70KB
```

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -88,6 +89,7 @@ private void executeHandler(ChannelHandlerContext ctx, int streamId, Route route
             .findFirst();
 
     if (handlerMapOptional.isPresent()) {
+      httpRequestMethodName = handlerMapOptional.get().keySet().asList().get(0).name();
```
jkinkead: On line 81 above, the method name is calculated.

You should be able to speed things up by calculating this once _before_ line 69, and saving the method name in the local.

jkinkead: (you'll speed things up by avoiding the calls here, but also by avoiding re-calculating the method in the inner loop)

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -108,6 +110,14 @@ private void executeHandler(ChannelHandlerContext ctx, int streamId, Route route
                   .handle(ctx.channel().attr(XrpcConstants.XRPC_REQUEST).get());
     }
 
+    // Check here for the case of an admin endpoint (eg /metrics, /health, and all others configured
+    // in
+    // Router.serveAdmin()); we do not track metrics for admin endpoints.
+    String meterName = MetricsUtil.getMeterNameForRoute(route, httpRequestMethodName);
+    if (xctx.getMetersByRoute().get(meterName) != null) {
```
jkinkead: Cache result of this expression:
```java
// too lazy to look up actual type
MeterType meter = xctx.getMetersByRoute().get(meterName);
if (meter != null) {
  meter.mark();
}
```

src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
```diff
@@ -0,0 +1,25 @@
...
+  protected static String getMeterNameForRoute(Route route, XHttpMethod httpMethod) {
+    return getMeterNameForRoute(route, httpMethod.name());
+  }
+
+  protected static String getMeterNameForRoute(Route route, String httpMethod) {
+    String method = Optional.ofNullable(httpMethod).orElse("ANY");
+
+    String routeIdentifier;
+    if (route != null && route.toString() != null) {
+      routeIdentifier = route.toString().replace('/', '.');
```
jkinkead: This should probably be cached as a property of the route, or in a map somewhere.

src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -98,6 +98,13 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                       .handle(xrpcRequest);
             }
 
+            // Check here for the case of an admin endpoint (eg /metrics, /health, and all others
+            // configured in
+            // Router.serveAdmin()); we do not track metrics for admin endpoints.
+            String meterName = MetricsUtil.getMeterNameForRoute(route, request.method().name());
+            if (xctx.getMetersByRoute().get(meterName) != null) {
```
jkinkead: Cache expression result as above.

src/main/java/com/nordstrom/xrpc/server/Router.java
andyday: yes i'd say making the names as short and sweet but unique is the way to go.

lducharme: great, removed them.

src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -108,6 +102,15 @@ private void executeHandler(ChannelHandlerContext ctx, int streamId, Route route
                   .handle(ctx.channel().attr(XrpcConstants.XRPC_REQUEST).get());
     }
 
+    // Check here for the case of an admin endpoint (eg /metrics, /health, and all others configured
+    // in Router.serveAdmin()); we do not track metrics for admin endpoints.
+    Meter meter =
```
xjdr: I'd like to see the functional Optional pattern here like on line 113

```java
Optional<Meter> routeMeter =
    Optional.ofNullable(xctx.getMetersByRoute()
          .get(MetricsUtil.getMeterNameForRoute(route, httpRequestMethod.name()));

routeMeter.ifPresent(Meter::mark);

lducharme: Sounds good, changed it in Http2Handler and UrlRouter.

src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
jkinkead: `Optional` is much better for readability, but it adds extra garbage for each request. I'm planning on running a benchmarking series at some point; it would be good to see how this impacts things.

xjdr:
Once you fix the merge conflicts, I am good to approve.

lducharme:
Merge conflicts now fixed (and addressed one more comment that I missed before).

andyday:
All of my remaining comments are unblocking...   The one about cyclomatic complexity is worth noting...

andyday:
you still need to fix the build issues.  CorsTest is failing...

src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
lducharme: Makes sense, I updated this.

However, I also added a route identifier map to cache the route identifiers per one of Jesse's comments, so hopefully the way I refactored it still makes sense and reduces complexity.

lducharme:
Build issues fixed! Thanks all for the help

PR #107, opened 2018-02-06
Multi-argument log with format string instead of using concatenation.

jkinkead:
More correct; nominally faster for `debug`-level logs. Probably nominally slower for non-`debug` logs. Perf testing was a wash (which triggers lots of error strings, but is at INFO level).


andyday:
👍 yes!  thank you...  wanted to do the same

PR #108, opened 2018-02-06
Restructure folders to support multiple demos helper modules

andyday:
- Moved existing demo under demos folder.  Both it and the pipeline demo serve as examples of how to include additional demos.
- Moved src under xrpc sub-folder.   This will allow for additional helper or axillary modules such as xrpc-testing, etc.


andyday:
@pdex you mean demos/people and demos/dino ??  yeah that was my plan.  also thinking about adding a hello world with simplest setup to be shown as getting started in readme...

pdex:
yeah that as well, maybe instead of people or dino we name the demo after what it shows off.

PR #112, opened 2018-02-07
Fix some cache issues with metrics names

jkinkead:
Name cache needs to be thread-safe.

Also, `Route` wasn't overriding `equals` or `hashCode`, so it's not going to work very well as a hash key.

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/server/http/Route.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/server/http/Route.java
@@ -127,4 +127,18 @@ public class Route {
   public String toString() {
     return originalPath;
   }
+
+  /** Two routes are considered equal if their original path specification was identical. */
+  @Override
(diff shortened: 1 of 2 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/server/http/Route.java
```diff
@@ -127,4 +127,18 @@ public boolean matches(String path) {
   public String toString() {
     return originalPath;
   }
+
+  /** Two routes are considered equal if their original path specification was identical. */
+  @Override
```
andyday: you can use lombok for these

jkinkead: But then I have to enumerate all of the fields in `Route` that I _don't_ want included in these calculations, and that's kind of icky.

I also lose the Javadoc on this.

PR #113, opened 2018-02-07
Move the two boolean paramters to `listenAndServe` into config values.

jkinkead:
Per discussion.

### DIFF
--- a/xrpc/src/test/java/com/nordstrom/xrpc/server/CorsTest.java
+++ b/xrpc/src/test/java/com/nordstrom/xrpc/server/CorsTest.java
@@ -31,6 +31,8 @@ class CorsTest {
   @BeforeEach
   void beforeEach() {
     config = ConfigFactory.load("test.conf").getConfig("xrpc");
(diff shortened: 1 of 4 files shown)


xrpc/src/test/java/com/nordstrom/xrpc/server/CorsTest.java
```diff
@@ -31,6 +31,8 @@
   @BeforeEach
   void beforeEach() {
     config = ConfigFactory.load("test.conf").getConfig("xrpc");
+    addConfigValue("serve_admin_routes", fromAnyRef(false));
+    addConfigValue("run_background_health_checks", fromAnyRef(false));
```
andyday: better to do this:

```java
config = ConfigFactory.load("test.conf").getConfig("xrpc")
  .withValue("serve_admin_routes", fromAnyRef(false))
  .withValue("run_background_health_checks", fromAnyRef(false));
```

jkinkead: Updated, I'll merge on green.

PR #114, opened 2018-02-07
Reorder issue template to be temporal.

jkinkead:
The current template confuses me greatly by a) not having a summary and b) having the steps to reproduce after the expected / actual outcomes.


andyday:
it would also be good to have headings for feature issues.  also examples or instructional text under each heading would be awesome....

jkinkead:
I'm going to leave out the instructional text for now, but I think it's a good idea.

PR #117, opened 2018-02-07
Inline config access.

jkinkead:
There is nothing special about these values that warrants their promotion to member variables.


andyday:
👍

PR #119, opened 2018-02-08
Health & metrics cleanup

### LINKED ISSUE #115: Health checks not registered unless background health checks
[The health registry function](https://github.com/Nordstrom/xrpc/blob/fe49c429ed8768d12faf59bdae8ea534ad844cd8/xrpc/src/main/java/com/nordstrom/xrpc/server/Router.java#L116-L120) only adds a check into a local map.

The checks are only registered when [the background health checker starts up](https://github.com/Nordstrom/xrpc/blob/fe49c429ed8768d12faf59bdae8ea534ad844cd8/xrpc/src/main/java/com/nordstrom/xrpc/server/Router.java#L129-L131), which can be disabled by a boolean flag.

These should _always_ be registered.

jkinkead:
Fixes #115 . It also fixes an unfiled bug where health checks are ran once before the handler lambda is created instead of once per invocation of `/health`.

This also uses a nonworker pool for health checks to avoid thread starvation. Pool size is resting-state zero by default.

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
@@ -181,8 +181,8 @@ public class AdminHandlers {
    */
   // CHECKSTYLE:ON
   public static Handler metricsHandler(MetricRegistry metrics, ObjectMapper mapper) {
-    Preconditions.checkState(metrics != null);
-    Preconditions.checkState(mapper != null);
+    Preconditions.checkArgument(metrics != null, "metrics may not be null");
+    Preconditions.checkArgument(mapper != null, "mapper may not be null");
     return xrpcRequest ->
         Recipes.newResponseOk(
             xrpcRequest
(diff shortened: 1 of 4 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -101,15 +100,13 @@ private void addResponseCodeMeters(XrpcConnectionContext.Builder contextBuilder)
...
       contextBuilder.meterByStatusCode(entry.getKey(), metricRegistry.meter(meterName));
     }
   }
 
-  public void addHealthCheck(String s, HealthCheck check) {
-    Preconditions.checkState(
-        !healthCheckMap.containsKey(s), "A Health Check by that name has already been registered");
-    healthCheckMap.put(s, check);
+  public void addHealthCheck(String name, HealthCheck check) {
+    healthCheckRegistry.register(name, check);
```
andyday: seems like this is the only functional change other than config and cleanup.  correct?

jkinkead: The PR description has a few others.

xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
```diff
@@ -181,8 +181,8 @@
    */
   // CHECKSTYLE:ON
   public static Handler metricsHandler(MetricRegistry metrics, ObjectMapper mapper) {
-    Preconditions.checkState(metrics != null);
-    Preconditions.checkState(mapper != null);
+    Preconditions.checkArgument(metrics != null, "metrics may not be null");
```
jkinkead: Functional change I didn't call out: This updates argument checks to use `checkArgument` instead of `checkState`. This throws `IllegalArgumentException` instead of `IllegalStateException`, which should be reserved for programmer errors ("I screwed up and this shouldn't happen ever").

xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
```diff
@@ -238,19 +238,18 @@ public static Handler gcHandler() {
...
-    return xrpcRequest ->
-        Recipes.newResponseOk(
-            xrpcRequest
-                .getAlloc()
-                .directBuffer()
-                .writeBytes(
-                    mapper.writerWithDefaultPrettyPrinter().writeValueAsBytes(healthChecks)),
-            Recipes.ContentType.Application_Json);
+    return xrpcRequest -> {
+      SortedMap<String, HealthCheck.Result> healthChecks = healthCheckRegistry.runHealthChecks();
```
jkinkead: Another functional change: This was being run once, _before_ the lambda was constructed, meaning that what was being served on `/health` was results from when the server first booted up, not real-time results.

xrpc/src/main/java/com/nordstrom/xrpc/server/Router.java
```diff
@@ -75,6 +73,7 @@ public Router(Config config) {
   public Router(XConfig config) {
     this.config = config;
     this.tls = new Tls(config.cert(), config.key());
+    this.healthCheckRegistry = new HealthCheckRegistry(config.asyncHealthCheckThreadCount());
```
jkinkead: Final functional change: This doesn't use the shared worker pool here for test running. This is a no-op unless you register health checks annotated with `Async`.

xrpc/src/main/java/com/nordstrom/xrpc/server/Router.java
jkinkead: Code annotated with others, now that I have time. :)

xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
andyday: good catch.

andyday:
looks good 👍

PR #106, opened 2018-02-06
Wip extract routes

pdex:

PR #120, opened 2018-02-09
New routes implementation

jkinkead:
This is split into two commits; the first adds a new routes implementation with unit tests, and the second replaces the current routes with the new routes. This builds on @pdex's work in #106 .

Design callouts:
* `Routes` is the builder class, but with a cleaner / friendlier name. It "compiles" to a `CompiledRoutes`, which is read-only (and thread-safe).
* `CompiledRoutes` handles (path, http method) -> (handler) mapping, as well as per-path metrics.
* `Router` is renamed `Server`.
* `XHttpMethod` and ANY handling no longer exists.

Functional changes:
* per-path metrics format looks like `routes.METHOD./my/path` instead of `route.method.my.path`.
* admin paths access is now tracked with metrics.
* path metrics are marked (counted) even when the handler throws an exception.
* routes may not be added after the server is started (after `listenAndServe` is invoked).

Implementation changes:
* HTTP/2 now passes the path handler through the context instead of re-matching paths.
* `Route` sorts naturally by `toString` instead of requiring a comparator.


xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -0,0 +1,127 @@
...
+   * @return the handler for the given path and method, if a handler matched. The handler returned
+   *     will set the match groups on the request when passed in, then forward the request to the
+   *     user-provided handler.
+   */
+  public Optional<Match> match(String path, HttpMethod method) {
+    for (RouteWithHandlers routeWithHandlers : routes) {
+      Route route = routeWithHandlers.getRoute();
+      Map<String, String> groups = route.groups(path);
+      if (groups != null) {
+        Handler handler = routeWithHandlers.getHandlers().get(method);
```
andyday: it we got a non-null groups but no method this would be a `405`  should handle appropriately here.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -155,35 +107,20 @@ public int onDataRead(
...
+        return processed;
+      } catch (IOException e) {
+        log.error("Error in handling Route", e);
+        // Error
+        ByteBuf buf = ctx.channel().alloc().directBuffer();
+        buf.writeBytes("Error executing endpoint".getBytes(XrpcConstants.DEFAULT_CHARSET));
+        writeResponse(ctx, streamId, HttpResponseStatus.INTERNAL_SERVER_ERROR, buf);
       }
     }
     return processed;
```
andyday: shoudn't this be `return 0;` ?

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -207,35 +145,41 @@ public void onHeadersRead(
       return;
     }
 
+    // After headers are read, we can determine if we have a route that handles the request. If
+    // there is no handler, we short-circuit 404 and drop and data. Otherwise, we'll pass through to
```
andyday: should be `and drop data`

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -207,35 +145,41 @@ public void onHeadersRead(
...
+    // the handler iff there is no data expected.
```
andyday: `the handler if there ...`

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -207,35 +145,41 @@ public void onHeadersRead(
...
+        try {
+          FullHttpResponse response = (FullHttpResponse) match.getHandler().handle(request);
+          sendResponse(ctx, response, streamId);
           return;
+        } catch (IOException e) {
+          log.error("Error in handling Route", e);
+          // Error
+          ByteBuf buf = channel.alloc().directBuffer();
+          buf.writeBytes("Error executing endpoint".getBytes(XrpcConstants.DEFAULT_CHARSET));
+          writeResponse(ctx, streamId, HttpResponseStatus.INTERNAL_SERVER_ERROR, buf);
```
andyday: if this is an internal server error we should `return;` here.  this also means the `else` below can be removed and its inner scope moved out...

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -207,35 +145,41 @@ public void onHeadersRead(
...
-          }
-        } else {
+    HttpMethod method = HttpMethod.valueOf(headers.method().toString());
+    Optional<CompiledRoutes.Match> matchOpt = xctx.getRoutes().match(path, method);
+    if (matchOpt.isPresent()) {
+      CompiledRoutes.Match match = matchOpt.get();
+      XrpcRequest request =
+          new XrpcRequest(headers, xctx.getMapper(), match.getGroups(), channel, streamId);
+      Optional<CharSequence> contentLength = Optional.ofNullable(headers.get(CONTENT_LENGTH));
+      if (!contentLength.isPresent()) {
```
andyday: won't the endOfStream in onData be called in this case?  if so could we not handle this there?

xrpc/src/main/java/com/nordstrom/xrpc/server/MetricsUtil.java
```diff
@@ -1,32 +0,0 @@
-package com.nordstrom.xrpc.server;
```
andyday: yes!

xrpc/src/main/java/com/nordstrom/xrpc/server/Routes.java
```diff
@@ -0,0 +1,154 @@
...
+   *     empty; or if there is already a POST handler for the route.
+   */
+  public Routes post(String route, Handler handler) {
+    return addRoute(route, handler, HttpMethod.POST);
+  }
+
+  /**
+   * Binds a handler for PUT requests to the given route.
+   *
+   * @throws IllegalArgumentException if either of the route or handler is null; if the route is
```
andyday: can remove `of` in `... either of the route ...` . same in all of these javadocs

xrpc/src/main/java/com/nordstrom/xrpc/server/Routes.java
```diff
@@ -0,0 +1,154 @@
...
+    Route route = Route.build(routePattern);
+
+    Map<HttpMethod, Handler> methods = routes.get(route);
+    if (methods == null) {
+      methods = new HashMap<>();
+      routes.put(route, methods);
+    }
+
+    // Verify that this method doesn't already exist.
+    if (methods.containsKey(method)) {
```
andyday: can use Preconditions.checkArgument here for consistency

xrpc/src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -64,65 +56,27 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
-            // in Router.serveAdmin()); we do not track metrics for admin endpoints.
-            Optional<Meter> routeMeter =
-                Optional.ofNullable(
-                    xctx.getMetersByRoute()
-                        .get(MetricsUtil.getMeterNameForRoute(route, request.method().name())));
-            routeMeter.ifPresent(Meter::mark);
+        HttpResponse resp = match.getHandler().handle(xrpcRequest);
 
-            xctx.getMetersByStatusCode().get(resp.status()).mark();
+        xctx.getMetersByStatusCode().get(resp.status()).mark();
```
andyday: ideally this should be done in the meteredHandler...

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -120,114 +124,11 @@ public void scheduleHealthChecks(
         () -> healthCheckRegistry.runHealthChecks(workerGroup), initialDelay, delay, timeUnit);
   }
 
-  /** Binds a handler for GET requests to the given route. */
```
andyday: it would be nice to have shortcut methods here for adding routes.   the default usage for hello world would be 

```java
Server server = new Server();
server.get("/hello-world", request -> request.okResponse("Hello World"));
server.listenAndServe();
```

BAM!  3 lines of code...

xrpc/src/test/java/com/nordstrom/xrpc/server/RoutesTest.java
```diff
@@ -0,0 +1,119 @@
...
+import com.google.common.collect.ImmutableMap;
+import io.netty.handler.codec.http.HttpMethod;
+import java.util.HashMap;
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+
+/** Tests for Routes and CompiledRoutes. */
+class RoutesTest {
+  @Test
+  public void addRoute_trivialMatch() throws Exception {
```
andyday: naming for these is not clear.  what's being tested here is Routes.get, Routes.compile, and CompiledRoutes.match.

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Gets the handler and matched groups for the given path and method.
+   *
+   * @return the handler for the given path and method, if a handler matched. The handler returned
+   *     will set the match groups on the request when passed in, then forward the request to the
+   *     user-provided handler.
+   */
+  public Optional<Match> match(String path, HttpMethod method) {
+    for (RouteWithHandlers routeWithHandlers : routes) {
```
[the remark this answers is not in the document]
jkinkead: I used `List` because it is a bit more natural (I only need iteration order of pairs), but I went ahead and made it a map anyway.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
jkinkead: I'm not sure, but it looks like yes.

But also: I think this whole implementation is broken if multiple DATA frames are received and should be reworked. I'd like to keep as-is for now to limit the already-large scope of changes.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
jkinkead: I have a TODO in the code I added to look into this. I'm going to leave this as-is (maintain HTTP/2 status quo).

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
jkinkead: Opting to maintain status quo for now, although I think you are correct.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
andyday: agreed.  maybe a TODO and issue to track this...

jkinkead: I'll file a catchall for HTTP/2 fixes and list these.

xrpc/src/main/java/com/nordstrom/xrpc/server/Routes.java
andyday: Both are grammatically correct, but ~~the~~ `of` is ~~considered~~ an unneeded word.  There's a school of thought that unneeded words ~~should~~ be removed to increase the simplicity and ~~therefore the~~ clarity of communication.   I've struck through all words that can be removed based on this argument.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
andyday: it's typical to expose the server.<route> methods for the default/getting-started usage and provide Routes for advanced usage.   it does add work for us but it makes it just a little bit simpler for users and anything we can do to make that happen is a good thing...

xrpc/src/main/java/com/nordstrom/xrpc/server/Routes.java
jkinkead: Citation needed . . . but I'll accept that it's probably fine.

andyday: http://sites.ieee.org/pcs/communication-resources-for-engineers/style/write-clearly-and-concisely/

although it needs more citation itself (https://en.wikipedia.org/wiki/Pleonasm)

https://www.testden.com/toeic/grammar-unnecessary-words.htm

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -0,0 +1,127 @@
...
+   * Returns compiled routes built from the given route map.
+   *
+   * @param metricRegistry the registry to generate per-(route,method) rate statistics in
+   */
+  public CompiledRoutes(
+      Map<Route, Map<HttpMethod, Handler>> rawRoutes, MetricRegistry metricRegistry) {
+    // Build a sorted list of the routes.
+    List<RouteWithHandlers> sortedRoutes = new ArrayList<>();
+    for (Route route : Sets.newTreeSet(rawRoutes.keySet())) {
+      ImmutableMap.Builder<HttpMethod, Handler> handlers = new ImmutableMap.Builder<>();
```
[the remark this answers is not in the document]
andyday: agreed on path handling inefficiencies.

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -98,30 +99,22 @@ public CompiledRoutes(
   }
 
   /** Container class for storing a route coupled with the handler implementations for it. */
+  @Value
   private static class RouteWithHandlers {
```
andyday: is this still used?  i don't think it is...

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -146,7 +146,7 @@ public void onHeadersRead(
     }
 
     // After headers are read, we can determine if we have a route that handles the request. If
-    // there is no handler, we short-circuit 404 and drop and data. Otherwise, we'll pass through to
+    // there is no handler, we short-circuit 404 and drop data. Otherwise, we'll pass through to
     // the handler iff there is no data expected.
     XrpcConnectionContext xctx = channel.attr(XrpcConstants.CONNECTION_CONTEXT).get();
     String path = getPathFromHeaders(headers);
```
andyday: what if contentLength is there but `0`?

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
andyday: i'm not sure iff is what is being done below nor does it make sense to me (at least) in this context.  please enlighten me...

andyday:
after conflicts have been resolved i'm good with this.  all of my recent comments are non-blocking.  the one about content length, I suspect is an extreme edge case and should be added to the http2 issue you opened at minimum.

xrpc/src/main/java/com/nordstrom/xrpc/server/Routes.java
jkinkead: I'm actually looking for the grammar rule that says that the "of" is superfluous. I can't find anything in that regard.

Extra words being removed is absolutely standard for technical writing. :)

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
jkinkead: Mostly I don't want to duplicate the helpers in both places. How do you feel about putting `Routes` behind an accessor, and having it only support low-level API calls? This is different from what @pdex originally had, but it keeps the API clean, and doesn't duplicate the `get`, `put`, etc helpers.

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
jkinkead: Nope! But I also deleted it in the last change. :)

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
jkinkead: Note added to tracking bug.

jkinkead:
I'm going to resolve conflicts and merge on green.

I'll refactor `Routes` in a separate PR, since it's a bunch of copy-paste and I wanted to isolate it.

I'll then look into the HTTP/2 issues.

PR #123, opened 2018-02-10
Demo testing

andyday:
added unit/api testing to people demo.


demos/people/src/test/java/com/nordstrom/xrpc/demos/people/PeopleTests.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class PeopleTests {
+  // don't use 8080 here as it may conflict with a running local app server
+  private static final String endpoint = "https:127.0.0.1:8014";
+  private final Application app = new Application(ConfigFactory.load("test.conf"));
+  private final OkHttpClient client = Http.unsafeClient();
+
+  @BeforeEach
```
lducharme: Just for my own understanding, what is the advantage to starting and stopping the app before each test vs starting it once for all tests?

andyday: not obvious here but its useful in situations where you want to change server configuration for each test.  given this I usually start with this approach and then only use `@Before` if there are performance issues with running tests.

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
```diff
@@ -22,9 +23,14 @@ public PeopleRoutes(Router router) {
...
     Handler getPerson =
         request -> {
-          String person = request.variable("person");
-          return request.okJsonResponse(
-              people.stream().anyMatch(p -> Objects.equals(p.name, person)));
+          String name = request.variable("person");
+          Optional<Person> person =
+              people.stream().filter(p -> Objects.equals(p.name, name)).findFirst();
+
+          if (person.isPresent()) {
```
jkinkead: Make `person` final and use `map` + `orElse` instead?

```java
final Optional<Person> person = ...;

// Rebuilds not found response with each request, but that's fine.
person.map(request::okJsonResponse).orElse(request.notFoundJsonResponse("Person Not Found"));
```

demos/people/src/test/java/com/nordstrom/xrpc/demos/people/PeopleTests.java
```diff
@@ -0,0 +1,86 @@
...
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class PeopleTests {
+  // don't use 8080 here as it may conflict with a running local app server
```
jkinkead: Ideally this would dynamically choose a port to avoid conflicts.

You should probably pull the port value from the config otherwise.

andyday: added to config.  and opened #125

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
andyday: map forces an IOException to be caught which makes this less a bit more complex to do.  leaving as is.

PR #124, opened 2018-02-12
Add a RouteBuilder mixin-style interface for things that build routes.

jkinkead:
I'm not sure that `Routes` _needs_ to implement the new interface, but it gets some extra features for free that way, since it already needs to implement the one abstract method.


jkinkead:
Class names swapped!

andyday:
👍 LGTM once you've resolved conflicts

Issue #121, opened 2018-02-09
Add per-handler timer metrics


jkinkead:
We have per-handler meters, and should add per-handler [Timers](http://metrics.dropwizard.io/4.0.0/manual/core.html#man-core-timers).

We should also come up with a good naming scheme for these, possibly fixing #104 first.

jkinkead:
Fixed in #127.

PR #127, opened 2018-02-13
Added per-handler timers 

lducharme:
Per issue [121](https://github.com/Nordstrom/xrpc/issues/121).

However, I'm unsure exactly where this timer should start and stop; should it stop timing after Handler.handle(request) is called? On channelInactive? Some other time?


xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -62,13 +64,17 @@ public CompiledRoutes(
 
         // Wrap the user-provided handler in one that tracks request rates.
         String metricName = MetricRegistry.name("routes", method.name(), route.toString());
+        String timerName = MetricRegistry.name("routeLatency", method.name(), route.toString());
```
andyday: These names need to have a common scheme.  Here are some ideas:

## Option 1
`routes.GET./path/to/my-service.requests`
`routes.GET./path/to/my-service.duration`

## Option 2

`routes.requests.GET./path/to/my-service`
`routes.duration.GET./path/to/my-service`

I prefer Option 1 as it organizes on the router first and then the type of metric.  @jkinkead I'd love to hear your thoughts on this as well.

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -62,13 +64,17 @@ public CompiledRoutes(
...
         final Meter meter = metricRegistry.meter(metricName);
+        final Timer timer = metricRegistry.timer(timerName);
         Handler meteredHandler =
             request -> {
               meter.mark();
-              // TODO(jkinkead): Add a timer here per https://github.com/Nordstrom/xrpc/issues/121
-              return userHandler.handle(request);
+              timer.time();
+              HttpResponse response = userHandler.handle(request);
+              timer.time().stop();
```
andyday: you can simplify this to:

```java
return timer.time(() -> userHandler.handle(request));
```

andyday: we really need to add unit tests for this....

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
andyday: we already have #103 to resolve metric naming so let's resolve this naming issue there...

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -62,13 +63,15 @@ public CompiledRoutes(
...
+        String timerName = MetricRegistry.name("routeLatency", method.name(), route.toString());
         final Handler userHandler = methodHandlerEntry.getValue();
         final Meter meter = metricRegistry.meter(metricName);
+        final Timer timer = metricRegistry.timer(timerName);
         Handler meteredHandler =
             request -> {
               meter.mark();
-              // TODO(jkinkead): Add a timer here per https://github.com/Nordstrom/xrpc/issues/121
-              return userHandler.handle(request);
+              timer.time();
```
andyday: this line should be removed

lducharme: thanks for catching this!

xrpc/src/main/java/com/nordstrom/xrpc/server/http/Recipes.java
```diff
@@ -58,9 +58,9 @@ public static ByteBuf unpooledBuffer(String payload) {
 
   // Request {{{
   public static FullHttpRequest newFullRequest(
-      HttpMethod method, String urlPath, ByteBuf payload, ContentType contentType) {
+    HttpMethod method, String urlPath, ByteBuf payload, ContentType contentType) {
```
andyday: make sure you've run ./gradlew goJV ...  these should not have changed.

xrpc/src/main/java/com/nordstrom/xrpc/server/Handler.java
```diff
@@ -17,10 +17,9 @@
 package com.nordstrom.xrpc.server;
 
 import io.netty.handler.codec.http.HttpResponse;
-import java.io.IOException;
 
 /** A handler for an HTTP route. */
 @FunctionalInterface
 public interface Handler {
-  HttpResponse handle(XrpcRequest xrpcRequest) throws IOException;
+  HttpResponse handle(XrpcRequest xrpcRequest) throws Exception;
```
jkinkead: This isn't part of the PR description and doesn't seem related. That said, it's a good change.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -115,7 +115,7 @@ public int onDataRead(
         FullHttpResponse response = (FullHttpResponse) handler.handle(request);
         sendResponse(ctx, response, streamId);
         return processed;
-      } catch (IOException e) {
+      } catch (Exception e) {
```
jkinkead: These try-catch blocks need to be more tightly-scoped now that they're catching just `Exception`.

xrpc/src/main/java/com/nordstrom/xrpc/server/Handler.java
lducharme: The reason I changed this is that when I updated the call to timer.time() to use lambda syntax (line 73 of CompiledRoutes currently), the code no longer compiled due to Handler.handle throwing Exception.

jkinkead: A cleaner solution to that would've been something like:

```java
try {
  // handle call
} catch (IOException|RuntimeException e) {
  // From child handler.
  throw ioe;
} catch (Exception e) {
  // Should be impossible.
  throw new IllegalStateException("unexpected exception", e);
}
```
This catches exceptions your Callable is expected to throw, and wraps unexpected ones in an unchecked exception.

lducharme: I see, this makes sense! I will change my code to do this and I'll revert the Http2Handler change.

A question about style- is it okay to leave this try-catch block in the lambda expression or is it preferred to move it to a lambda wrapper?

jkinkead: My opinion is that it's fine to put anything into a lambda that you'd put in other code blocks.

andyday:
👍 LGTM

PR #130, opened 2018-02-14
Use a thread-safe PeopleRoutes implementation.

jkinkead:
Tiny fix to make this a better example class.

### DIFF
--- a/demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
+++ b/demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
@@ -2,14 +2,14 @@ package com.nordstrom.xrpc.demos.people;
 
 import com.nordstrom.xrpc.server.Handler;
 import com.nordstrom.xrpc.server.Routes;
-import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
```diff
@@ -2,14 +2,14 @@
...
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.CopyOnWriteArrayList;
 import lombok.Value;
 
 public class PeopleRoutes {
-  private final List<Person> people = new ArrayList<>();
+  private final List<Person> people = new CopyOnWriteArrayList<>();
```
andyday: not blocking but maybe this should be concurrent version of a set...   since its demo probably not worth thinking about too much

jkinkead: Yeah, it's a fairly terrible demo. It should really use Jackson with more than one field, not just take the whole request body as the name ... :)

PR #136, opened 2018-02-14
Exception Handling Part 1

andyday:
This is part 1 of a 2 part feature.  The goal is to provide conversion of Exceptions to proper HTTP Responses.  This PR sets the stage by allowing for the registering of an ExceptionHandler with the fall back to a DefaultExceptionHandler that does nothing more than re-throw the exception.  In part 2 the DefaultExceptionHandler will look for specific types of exceptions and return appropriate Responses based on a reasonable set of rules...


xrpc/src/main/java/com/nordstrom/xrpc/server/ExceptionHandler.java
```diff
@@ -0,0 +1,9 @@
+package com.nordstrom.xrpc.server;
+
+import io.netty.handler.codec.http.HttpResponse;
+
+/** A Handler for mapping exceptions to Http Responses. */
+@FunctionalInterface
+public interface ExceptionHandler {
+  HttpResponse handle(XrpcRequest request, Exception exception) throws Exception;
```
jkinkead: It's not clear under what conditions this should throw an exception. Document? Ideally, this should _never_ throw exceptions (it's meant as a fallback).

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -109,6 +109,10 @@ private void addResponseCodeMeters(XrpcConnectionContext.Builder contextBuilder)
     }
   }
 
+  public void registerExceptionHandler(ExceptionHandler handler) {
```
jkinkead: Document? In particular, it's unclear if this overwrites the only exception handler, or appends an exception handler to a list.

xrpc/src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -54,12 +53,16 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
 
-      ObjectMapper mapper = xctx.getMapper();
       XrpcRequest xrpcRequest =
           new XrpcRequest(request, xctx.getMapper(), match.getGroups(), ctx.channel());
       xrpcRequest.setData(request.content());
 
-      HttpResponse resp = match.getHandler().handle(xrpcRequest);
+      HttpResponse resp;
+      try {
+        resp = match.getHandler().handle(xrpcRequest);
```
jkinkead: You should probably scope the `try` block to only the `handle` call. We don't want errors in our `getHandler()` method to be sent to the user.

xrpc/src/main/java/com/nordstrom/xrpc/server/DefaultExceptionHandler.java
```diff
@@ -0,0 +1,12 @@
+package com.nordstrom.xrpc.server;
+
+import io.netty.handler.codec.http.HttpResponse;
+
+public class DefaultExceptionHandler implements ExceptionHandler {
+  @Override
+  public HttpResponse handle(XrpcRequest request, Exception exception) throws Exception {
+    // TODO (AD): Use configured response content/type and response specific exceptions to default
+    // to meaningful responses here
+    throw exception;
```
jkinkead: I would omit the default handler if all it's doing is rethrowing, since that can be implemented in code (`if handler == null throw e`).

xrpc/src/main/java/com/nordstrom/xrpc/server/UrlRouter.java
```diff
@@ -54,12 +53,16 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
       CompiledRoutes.Match match = xctx.getRoutes().match(path, request.method());
 
-      ObjectMapper mapper = xctx.getMapper();
       XrpcRequest xrpcRequest =
           new XrpcRequest(request, xctx.getMapper(), match.getGroups(), ctx.channel());
       xrpcRequest.setData(request.content());
 
-      HttpResponse resp = match.getHandler().handle(xrpcRequest);
+      HttpResponse resp;
+      try {
```
jkinkead: `Handler.handle` is invoked twice in `Http2Handler` as well.

xrpc/src/main/java/com/nordstrom/xrpc/server/DefaultExceptionHandler.java
andyday: at this point it is a placeholder.  in part 2 it will do more than just rethrow

andyday:
@jkinkead i've moved the logic to compiled routes and added some clarifying documentation...

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
```diff
@@ -68,20 +68,26 @@ public CompiledRoutes(
...
-                // Should be impossible.
-                throw new IllegalStateException("unexpected exception", e);
+                try {
+                  return request.getConnectionContext().getExceptionHandler().handle(request, e);
+                } catch (IOException | RuntimeException ioe) {
+                  // From child handler.
+                  throw ioe;
+                } catch (Exception ee) {
+                  // Should be impossible.
+                  throw new IllegalStateException("unexpected exception", ee);
```
jkinkead: You've inverted the wrapping here; you should pass through only the expected child exceptions:
```java
try {
  return timer.time(() -> userHandler.handle(request));
} catch (IOException | RuntimeException handlerE) {
  try {
    return request.getConnectionContext().getExceptionHandler().handle(request, handlerE);
  } catch (Exception ee) {
    // TODO: Handle gracefully. This should probably log and return a generic 500.
  }
} catch (Exception e) {
  // Should be impossible.
  throw new IllegalStateException("unexpected exception", ee);
}
```

xrpc/src/main/java/com/nordstrom/xrpc/server/ExceptionHandler.java
```diff
@@ -0,0 +1,19 @@
...
+import io.netty.handler.codec.http.HttpResponse;
+
+/** A Handler for mapping exceptions to Http Responses. */
+@FunctionalInterface
+public interface ExceptionHandler {
+  /**
+   * Handle exception. This is used to handle an exception thrown by the registered Route Handler.
+   * It converts the Exception to an appropriate HttpResponse.
+   *
+   * @param request Request for which this Exception was generated
```
jkinkead: Lower-case first word of non-full-sentence Javadoc for `@param`.

xrpc/src/main/java/com/nordstrom/xrpc/server/ExceptionHandler.java
```diff
@@ -0,0 +1,19 @@
...
+   * @param exception Thrown exception
+   * @return HttpResponse appropriate for this Exception.
```
jkinkead: Omit terminal period for sentence fragment `@return`

xrpc/src/main/java/com/nordstrom/xrpc/server/ExceptionHandler.java
```diff
@@ -0,0 +1,19 @@
...
+   * @throws Exception Exception for any unhandled Exception passed in or during the conversion
```
jkinkead: I don't think we want to allow this; I don't see this as a valid use-case. Handlers should be in charge of generating a valid response for all requests.

If this throws, we should be catching it and returning a generic 500 response.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -109,6 +109,16 @@ private void addResponseCodeMeters(XrpcConnectionContext.Builder contextBuilder)
     }
   }
 
+  /**
+   * Set the customer exception handler. This handler is called to handle any exceptions thrown from
```
jkinkead: "custom" instead of "customer"?

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -109,6 +109,16 @@ private void addResponseCodeMeters(XrpcConnectionContext.Builder contextBuilder)
...
+   * route handlers. This replaces any other exception handlers previously set.
+   *
+   * @param handler Exception Handler
```
jkinkead: Omit this per google style guide? (omit `@param` if it's already covered by existing doc).

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -44,15 +43,14 @@
...
-  private final ObjectMapper mapper;
-
   /** The request to handle. */
   @Getter private final FullHttpRequest h1Request;
 
   @Getter private final Http2Headers h2Headers;
   @Getter private final EventLoop eventLoop;
   @Getter private final Channel upstreamChannel;
   @Getter private final ByteBufAllocator alloc;
+  @Getter private final XrpcConnectionContext connectionContext;
```
jkinkead: I _think_ we should make this package:
```java
import lombok.AccessLevel;
...
@Getter(AccessLevel.PACKAGE) private final XrpcConnectionContext connectionContext;
```
I don't believe that users will ever need to interact with this. It's currently safe for them to interact with, but it might not always remain so, and it's cluttering up the API at the moment.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -63,10 +61,13 @@
   private ByteBuf data;
 
   public XrpcRequest(
-      FullHttpRequest request, ObjectMapper mapper, Map<String, String> groups, Channel channel) {
+      FullHttpRequest request,
+      XrpcConnectionContext ctx,
```
jkinkead: Match parameter to member name?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -76,13 +77,13 @@ public XrpcRequest(
 
   public XrpcRequest(
       Http2Headers headers,
-      ObjectMapper mapper,
+      XrpcConnectionContext ctx,
```
jkinkead: Match parameter to member name?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -222,7 +223,7 @@ public FullHttpResponse badRequestJsonResponse(Object body) throws IOException {
   private ByteBuf bodyToByteBuf(Object body) throws IOException {
     ByteBuf buf = getAlloc().directBuffer();
     OutputStream stream = new ByteBufOutputStream(buf);
-    mapper.writeValue(stream, body);
+    this.connectionContext.getMapper().writeValue(stream, body);
```
jkinkead: Remove `this` for consistency with the rest of the file?

xrpc/src/main/java/com/nordstrom/xrpc/server/CompiledRoutes.java
andyday: that's what I get for doing this during a design review

andyday: actually i didn't invert it.  i first want to pass any Exception to the exception handler and if any exception comes from that I want to split into more granular exception handling

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: this could probably be said for some of the other XrpcRequest getters as well.

andyday:
Think I've addressed all your comments...

xrpc/src/main/java/com/nordstrom/xrpc/XrpcConstants.java
```diff
@@ -49,6 +49,9 @@
           .writeBytes(
               "This response is being send due to too many requests being sent to the server"
                   .getBytes(DEFAULT_CHARSET));
+
+  public static final byte[] INTERNAL_SERVER_ERROR_RESONSE =
+      "Internal Server Error".getBytes(DEFAULT_CHARSET);
```
jkinkead: This just makes line 50's response look too wordy . . . :)

xrpc/src/main/java/com/nordstrom/xrpc/server/DefaultExceptionHandler.java
```diff
@@ -0,0 +1,22 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class DefaultExceptionHandler implements ExceptionHandler {
+  @Override
+  public HttpResponse handle(XrpcRequest request, Exception exception) {
+    // TODO (AD): Use configured response content/type and response specific exceptions to default
+    // to meaningful responses here
+    log.error("Handler Exception: {}", exception.toString());
```
jkinkead: This should probably log the stack trace, for debuggability:
```java
log.error("Handler Exception:", exception);
```

xrpc/src/main/java/com/nordstrom/xrpc/XrpcConstants.java
andyday: i shortened the one on line 50 as it is the outlier.   a similar comment is coming on your PR that changes that line.

xrpc/src/main/java/com/nordstrom/xrpc/server/DefaultExceptionHandler.java
andyday: doh.   in C# toString() does the trick.   good catch.

xrpc/src/main/java/com/nordstrom/xrpc/XrpcConstants.java
jkinkead: 👍

PR #138, opened 2018-02-14
Admin handler cleanup

jkinkead:
This splits the handlers into two clear categories, and disables the "unsafe" handlers by default. I don't expect people to read the fine-print on these being dangerous.

This also:
* registers the GC handler properly
* updates the shutdown handler to call `shutdown()` inside the lambda instead of on lambda creation
* prevents `this`-leakage from Server constructor by registering admin paths on serve

FYI @xjdr .

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
@@ -267,10 +267,15 @@ public class AdminHandlers {
 
   /**
    * Registers informational admin routes. These may contain sensitive information, and should have
-   * sensible access controls imposed on them. Routes are: '/info': exposes version number, git
-   * commit number, etc '/metrics': returns the metrics reporters in JSON format '/health': exposes
-   * a summary of downstream health checks '/ping': responds with a 200-OK status code and the text
-   * 'PONG' '/ready': expose a Kubernetes or ELB specific healthcheck for liveliness
+   * sensible access controls imposed on them. Routes are:
+   *
+   * <ul>
+   *   <li>'/info': exposes version number, git commit number, etc
+   *   <li>'/metrics': returns the metrics reporters in JSON format
+   *   <li>'/health': exposes a summary of downstream health checks
+   *   <li>'/ping': responds with a 200-OK status code and the text 'PONG'
+   *   <li>'/ready': expose a Kubernetes or ELB specific healthcheck for liveliness
+   * </ul>
    */
   static void registerInfoAdminRoutes(Server server) {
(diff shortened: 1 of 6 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
```diff
@@ -267,10 +267,15 @@ public static Handler getKillHandler(Server server) {
...
   /**
    * Registers informational admin routes. These may contain sensitive information, and should have
-   * sensible access controls imposed on them. Routes are: '/info': exposes version number, git
-   * commit number, etc '/metrics': returns the metrics reporters in JSON format '/health': exposes
-   * a summary of downstream health checks '/ping': responds with a 200-OK status code and the text
-   * 'PONG' '/ready': expose a Kubernetes or ELB specific healthcheck for liveliness
+   * sensible access controls imposed on them. Routes are:
+   *
+   * <ul>
+   *   <li>'/info': exposes version number, git commit number, etc
```
andyday: i find it odd that no closing tag is needed...  javadoc must be conforming to a HTML 4.01 Transitional schema...

xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
```diff
@@ -257,17 +251,60 @@ public static Handler healthCheckHandler(
...
+   *   <li>'/ping': responds with a 200-OK status code and the text 'PONG'
+   *   <li>'/ready': expose a Kubernetes or ELB specific healthcheck for liveliness
+   * </ul>
+   */
+  static void registerInfoAdminRoutes(Server server) {
+    MetricsModule metricsModule = new MetricsModule(TimeUnit.SECONDS, TimeUnit.MILLISECONDS, true);
+    // TODO(jkinkead): This should optionally use a custom mapper.
+    ObjectMapper metricsMapper = new ObjectMapper().registerModule(metricsModule);
+    server.get(
+        "/metrics", AdminHandlers.getMetricsHandler(server.getMetricRegistry(), metricsMapper));
```
andyday: I think the admin routes should be moved to the AdminHandlers.  it could be renamed to something more appropriate, but it would be good to have the routes and the handling in the same place.

xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
jkinkead: HTML 5 doesn't require list items to be terminated, either. It makes me a little sad, but I thought it was the accepted style...

xrpc/src/main/java/com/nordstrom/xrpc/server/AdminHandlers.java
jkinkead: I think that's what I'm doing here?

PR #139, opened 2018-02-16
Enable Auto Port Selection for testing

### LINKED ISSUE #125: Dynamically generate test server port.
For unit testing APIs, run Server in test mode with automatically generated port number to avoid port conflicts.  This port number should be publicly exposed so that tests can build and endpoint with this port.  See PR #123

andyday:
This fixes #125

### DIFF
--- a/demos/people/src/main/java/com/nordstrom/xrpc/demos/people/Application.java
+++ b/demos/people/src/main/java/com/nordstrom/xrpc/demos/people/Application.java
@@ -21,26 +21,16 @@ import com.nordstrom.xrpc.server.Server;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import java.io.IOException;
-import lombok.Getter;
-import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class Application {
-  @Getter
-  @Accessors(fluent = true)
-  private final Server server;
-
-  public Application(Config config) {
-    // Build your server. This overrides the default configuration with values from
-    // src/main/resources/demo.conf.
-    this.server = new Server(config);
-
+  public static void configure(Server server) {
     // Add handlers for /people routes
     new PeopleRoutes(server);
 
     // Add a service specific health check
-    this.server.addHealthCheck(
+    server.addHealthCheck(
         "simple",
         new HealthCheck() {
           @Override
@@ -51,28 +41,20 @@ public class Application {
         });
   }
 
-  public void start() throws IOException {
-    server.listenAndServe();
-  }
-
-  public void stop() {
-    server.shutdown();
-  }
-
   public static void main(String[] args) {
     // Load application config from jar resources. The 'load' method below also allows supports
     // overrides from environment variables.
     Config config = ConfigFactory.load("demo.conf");
-
-    Application app = new Application(config);
+    Server server = new Server(config);
+    Application.configure(server);
 
     try {
       // Fire away
-      app.start();
+      server.listenAndServe();
     } catch (IOException e) {
       log.error("Failed to start people server", e);
     }
 
-    Runtime.getRuntime().addShutdownHook(new Thread(app::stop));
+    Runtime.getRuntime().addShutdownHook(new Thread(server::shutdown));
   }
 }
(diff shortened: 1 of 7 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -40,18 +40,21 @@
...
 @Slf4j
+@Accessors(fluent = true)
 public class Server implements Routes {
   private final XConfig config;
   private final Tls tls;
   private final XrpcConnectionContext.Builder contextBuilder;
   private final MetricRegistry metricRegistry = new MetricRegistry();
   @Getter private final RouteBuilder routeBuilder = new RouteBuilder();
 
+  @Getter private int port;
```
jkinkead: I think it's confusing that this doesn't match the configured port under some conditions.

What about having a `getActualPort` method that gets the port dynamically from the channel?

Alternately, documenting when this gets initialized and renaming to `actualPort` would be fine.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -40,18 +40,21 @@
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import lombok.Getter;
+import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
 import org.slf4j.LoggerFactory;
 
 /** An xprc server. */
 @Slf4j
+@Accessors(fluent = true)
```
jkinkead: I don't think we should add this here without adding it consistently across the API (like in `XrpcRequest` and other endpoints as well).

I'm fine with whatever, but we should try to have our public API be consistent.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
andyday: I've added javadocs to clarify this.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
andyday: leaving this as fluent for now with the plan to change all to fluent in another pr.

jkinkead: Sounds great!

andyday: see #142

andyday:
@jkinkead ready for re-review

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -40,18 +40,26 @@
...
 public class Server implements Routes {
   private final XConfig config;
   private final Tls tls;
   private final XrpcConnectionContext.Builder contextBuilder;
   private final MetricRegistry metricRegistry = new MetricRegistry();
   @Getter private final RouteBuilder routeBuilder = new RouteBuilder();
 
+  /**
+   * Actual port on which this server is running. This will differ from the configured port of it is
```
jkinkead: I really think the name should reflect this (`actualPort`). You should also be clear when this gets initialized - and may also want to use an obviously-bad placeholder, like `-1`, as an initial value.

andyday: the main use of this is in testing.  all the tests need to no is which port the server is running on.  to add the word actual adds to the confusion in my view

andyday:
think i've got the port stuff squared.  let me know.

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/Application.java
```diff
@@ -25,18 +25,12 @@
...
 @Slf4j
 public class Application {
-  private final Server server;
-
-  public Application(Config config) {
-    // Build your server. This overrides the default configuration with values from
-    // src/main/resources/demo.conf.
-    this.server = new Server(config);
-
+  public static void configure(Server server) {
```
jkinkead: This seems like a nice pattern.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -222,6 +253,13 @@ public void listenAndServe() throws IOException {
     }
 
     channel = future.channel();
+    InetSocketAddress actualAddress = (InetSocketAddress) channel.localAddress();
+    log.info("Listening at {}", actualAddress.getAddress().getCanonicalHostName());
+  }
+
+  public String localEndpoint() {
```
jkinkead: Maybe add a `@throws NullPointerException if called before listenAndServe`?

andyday: i'll add that on the next pr.

PR #140, opened 2018-02-16
Disable HTTP2 frame logging by default.

jkinkead:
Trying not to clutter up the pending http2 branch.

This disables some noisy debug logging that's currently on by default.

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/server/Http2HandlerBuilder.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/server/Http2HandlerBuilder.java
@@ -22,19 +22,19 @@ import io.netty.handler.codec.http2.Http2ConnectionEncoder;
 import io.netty.handler.codec.http2.Http2FrameLogger;
 import io.netty.handler.codec.http2.Http2Settings;
 import io.netty.handler.logging.LogLevel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class Http2HandlerBuilder
     extends AbstractHttp2ConnectionHandlerBuilder<Http2Handler, Http2HandlerBuilder> {
 
-  /**
-   * Builds a new builder with the given configuration values.
-   *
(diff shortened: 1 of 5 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/server/Http2HandlerBuilder.java
```diff
@@ -26,10 +26,16 @@
...
-    frameLogger(logger);
+  /**
+   * Builds a new builder with the given configuration values.
+   *
+   * @param logHttp2Frames if true, log each HTTP2 frame as it comes in. Should be false for
+   *     production deploys.
+   */
+  public Http2HandlerBuilder(boolean logHttp2Frames) {
+    if (logHttp2Frames) {
+      frameLogger(new Http2FrameLogger(LogLevel.INFO, Http2Handler.class.getName() + ".frames"));
```
andyday: Frame logging should be DEBUG level.

jkinkead: This was  conscious choice.

My logic here was: Since this is only meant to be used for local debugging, and requires a config change to turn on, requiring a *second* config change to increase the log level seems gratuitous.

What do you think?

andyday: currently when you have level=DEBUG, for http1 the "frames" are logged.  maybe what we do here is use level=DEBUG to turn this on

pdex: @jkinkead why not check that the logger is enabled?

jkinkead: Gating on DEBUG enabled is a great idea; done!

jkinkead:
Updated to be DEBUG and enabled based on logging configuration.

andyday:
👍 LGTM

PR #149, opened 2018-02-21
Adds benchmarking

spenserpothier:
* Adds JMH plugin
* Adds a simple benchmark for the CompiledRoutes.match method

Benchmarks can be run using `./gradlew clean jmh`
This first benchmark takes about 15 minutes to run and give an output similar to
the following, calculating the average time and throughput for the match method:

```
// Run complete. Total time: 00:13:47

Benchmark              Mode  Cnt  Score   Error   Units
JmhRouter.basicMatch  thrpt  200  9.105 ± 0.409  ops/us
JmhRouter.basicMatch   avgt  200  0.127 ± 0.011   us/op
```

### DIFF
--- a/xrpc/src/jmh/java/com/nordstrom/xrpc/server/JmhCompiledRoutesBenchmark.java
+++ b/xrpc/src/jmh/java/com/nordstrom/xrpc/server/JmhCompiledRoutesBenchmark.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2018 Nordstrom, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nordstrom.xrpc.server;
+
+import com.codahale.metrics.MetricRegistry;
+import io.netty.handler.codec.http.HttpMethod;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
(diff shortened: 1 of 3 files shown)


xrpc/src/jmh/java/com/nordstrom/xrpc/server/JmhCompiledRoutesBenchmark.java
```diff
@@ -0,0 +1,50 @@
...
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+@State(Scope.Thread)
+public class JmhCompiledRoutesBenchmark {
+  CompiledRoutes compiledRoutes;
+
+  @Setup
```
jkinkead: It would be better to have this be a fixture class, but it's not a big deal - we can update this later.

(We'll want to have multiple route configurations, testing how performance scales).

xrpc/src/jmh/java/com/nordstrom/xrpc/server/JmhCompiledRoutesBenchmark.java
```diff
@@ -0,0 +1,50 @@
...
+  public void setupRouter() {
+    Handler handler = request -> null;
+    MetricRegistry metricRegistry = new MetricRegistry();
+    RouteBuilder routes = new RouteBuilder();
+    routes.get("/foo", handler);
+    compiledRoutes = routes.compile(metricRegistry);
+  }
+
+  @BenchmarkMode({Mode.AverageTime, Mode.Throughput})
```
andyday: i'd like to find a setting that will allow this to run in 5mins or less so that we can efficiently run it in the pipeline...

spenserpothier: I don't think "running it faster" should be a concern during a benchmark. The default settings on JMH are 20 warmup iterations for the JVM to optimize, followed by 20 iterations of data collection. 

I do agree this should be added to the travis pipeline, but I don't think it should be a blocking stage of the build process. I have created #150 for tracking and discussion about how we want to handle this.

jkinkead: Yeah, trying to get this to run quickly seems like a bad idea. The whole point is that it runs for a long time in order to gather valid statistics...

PR #158, opened 2018-02-28
content-type based encoding/decoding

andyday:


xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public <T> HttpResponse internalServerError(T body) throws IOException {
+    return create(HttpResponseStatus.INTERNAL_SERVER_ERROR, body);
+  }
+
+  public <T> HttpResponse create(HttpResponseStatus status, T body) throws IOException {
+    final Encoder encoder =
+        request.connectionContext().encoders().acceptedEncoder(request.acceptHeader());
+    ByteBuf buf = request.byteBuf();
+    return create(status, encoder.encode(buf, body), encoder.contentType());
```
jkinkead: This should use the charset in both the `encode` step and in generating the response's content-type header.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public HttpResponse create(
+      HttpResponseStatus status,
+      ByteBuf body,
+      CharSequence contentType,
+      Map<String, String> customHeaders) {
+    FullHttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status, body);
+
+    if (customHeaders != null) {
+      for (Map.Entry<String, String> entry : customHeaders.entrySet()) {
```
jkinkead: `response.headers().set(customHeaders.entrySet())` instead?

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,82 @@
...
+  public HttpResponse create(HttpResponseStatus status, ByteBuf body, CharSequence contentType) {
+    return create(status, body, contentType, null);
+  }
+
+  public HttpResponse create(
+      HttpResponseStatus status,
+      ByteBuf body,
+      CharSequence contentType,
+      Map<String, String> customHeaders) {
+    FullHttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status, body);
```
jkinkead: This should only use `FullHttpResponse` if the body is non-empty.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,82 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpVersion;
+import java.io.IOException;
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@AllArgsConstructor
+public class ResponseFactory {
```
jkinkead: You could make this a mixin with a `getRequest` abstract field. This would let you pull it in to `XrpcRequest` directly.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
```diff
@@ -98,11 +108,27 @@ private Server(XConfig config, int port) {
     this.tls = new Tls(config.cert(), config.key());
     this.healthCheckRegistry = new HealthCheckRegistry(config.asyncHealthCheckThreadCount());
 
+    ObjectMapper mapper =
+        new ObjectMapper().registerModule(new ParameterNamesModule(JsonCreator.Mode.PROPERTIES));
```
jkinkead: What does this module do beyond the basics?

Maybe document?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -16,30 +16,21 @@
...
 import io.netty.handler.codec.http2.Http2Headers;
-import io.netty.handler.codec.http2.HttpConversionUtil;
-import java.io.IOException;
-import java.io.OutputStream;
 import java.nio.charset.Charset;
-import java.util.Collections;
 import java.util.Map;
 import lombok.AccessLevel;
 import lombok.Getter;
+import lombok.SneakyThrows;
```
jkinkead: Please don't use this. It makes checked exceptions impossible to catch.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -141,37 +134,42 @@ int addData(ByteBuf dataFrame) {
     return h2Data.readableBytes();
   }
 
+  @SneakyThrows
+  public <T> T body(Class<T> clazz) {
```
jkinkead: Maybe `bodyAs` instead of `body`?

You should also document that this consumes the request body, and can't be called repeatedly.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -141,37 +134,42 @@ int addData(ByteBuf dataFrame) {
...
     return h2Data;
   }
 
   /**
-   * Returns a new string representing the request data, decoded using the appropriate charset. This
+   * Returns a new string representing the request body, decoded using the appropriate charset. This
    * does NOT consume or mutate the underlying data buffer.
    */
-  public String getDataAsString() {
+  public String bodyText() {
```
jkinkead: `bodyAsString`?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -181,64 +179,11 @@ public CharSequence getHeader(CharSequence name) {
...
+  public CharSequence acceptHeader() {
+    return header(HttpHeaderNames.ACCEPT);
   }
 
-  private ByteBuf encodeJsonBody(Object body) throws IOException {
-    ByteBuf buf = alloc().directBuffer();
-    OutputStream stream = new ByteBufOutputStream(buf);
-    connectionContext.mapper().writeValue(stream, body);
-    return buf;
+  public CharSequence contentType() {
```
jkinkead: `contentTypeHeader` to match the new `acceptHeader` method name?

xrpc/src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -81,6 +81,14 @@ ip_black_list = []
 enable_white_list = false
 ip_white_list = []
 
+# Default Content-Type
+# This is used when no Accept or Content-Type header is given or the Accept header or Content-Type header
```
jkinkead: Include what the value is used for (to look up an encoder or decoder)? That's the part that's non-obvious.

xrpc/src/test/java/com/nordstrom/xrpc/EncodingTest.java
```diff
@@ -0,0 +1,91 @@
...
+  void testJsonEncoding() throws IOException {
+    Response response =
+        client
+            .newCall(
+                new Request.Builder()
+                    .url(endpoint + "/person")
+                    .post(
+                        RequestBody.create(
+                            MediaType.parse("application/json"), "{\"name\":\"bob\"}"))
+                    .header("Content-Type", "application/json")
```
jkinkead: You should include a test with charset specified.

xrpc/src/test/java/com/nordstrom/xrpc/EncodingTest.java
```diff
@@ -0,0 +1,91 @@
...
+  @Test
+  void testJsonDecoding() throws IOException {
+    person = new Person("bob");
+    Response response =
+        client
+            .newCall(
+                new Request.Builder()
+                    .url(endpoint + "/person")
+                    .get()
+                    .header("Accept", "application/json; text/plain")
```
jkinkead: You should include tests with charset set, and tests for wildcard types & subtypes.

xrpc/src/test/java/com/nordstrom/xrpc/EncodingTest.java
```diff
@@ -0,0 +1,91 @@
...
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class EncodingTest {
```
jkinkead: It would be nice to see unit tests of the Encoders & Decoders lookup logic.

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
```diff
@@ -40,8 +41,12 @@ public PeopleRoutes(Routes routes) {
   }
 
   // Application POJO for use in request / response.
-  @Value
   public static class Person {
-    private String name;
+    @Getter private String name;
+
+    @ConstructorProperties("name")
```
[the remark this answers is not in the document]
andyday: most json libraries use bean attributes so this works for Jackson and others.  the demo should show not only how to use xrpc but provide best practices as well.  the best practice here is avoid coupling to specific format types or implementations of serializers.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
```diff
@@ -0,0 +1,87 @@
...
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/** Holds a set of Decoders each registered to do decoding for a given content type. */
+@Slf4j
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class Decoders {
+  private final Decoder defaultDecoder;
+  private final ImmutableMap<CharSequence, Decoder> decoders;
```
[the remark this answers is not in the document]
andyday: that was the original reason for not using CharSequence but I wanted to try again without sorted map.  i would have expected it to fail unit tests unless they are missing this case somehow

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
```diff
@@ -0,0 +1,87 @@
...
+   *
+   * @param contentType content type header value. If null or not registered the default decoder
+   *     will be returned
+   * @return best Decoder for the given content type header
+   */
+  public Decoder decoder(CharSequence contentType) {
+    if (contentType == null) {
+      return defaultDecoder;
+    }
+    // TODO (AD): Handle charset see:
```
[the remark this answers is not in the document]
andyday: yes that's what it means...  will address.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public Encoder acceptedEncoder(CharSequence accept) {
+    if (accept == null) {
+      return defaultEncoder;
+    }
+
+    // TODO (AD): Consider LRU cache of accepts to encoders.
+    String[] contentTypes = ACCEPT_SPLIT_PATTERN.split(accept);
+    for (String contentType : contentTypes) {
+      // TODO (AD): Handle q-factor weighting see:
```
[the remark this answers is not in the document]
andyday: wildcard types can be handled by registering the encoder for that type for instance `application/*` could be registered to use JsonEncoder...  currently I'm just relying on falling back on default encoder.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
```diff
@@ -0,0 +1,94 @@
...
+      // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
+      Encoder encoder = encoders.get(contentType);
```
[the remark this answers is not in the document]
andyday: yes.  was planning on doing this in a future PR.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
```diff
@@ -0,0 +1,94 @@
...
+      if (encoder != null) {
+        return encoder;
+      }
+    }
+    return defaultEncoder;
+  }
+
+  /** Get Encoders builder. */
```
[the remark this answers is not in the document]
andyday: it can't be generated without moving the defaultEncoder lookup to a custom constructor on Encoders.  started with generated and worked my way to this as a simpler approach.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
```diff
@@ -0,0 +1,87 @@
...
+  private final ImmutableMap<CharSequence, Decoder> decoders;
+
+  /**
+   * Find a Decoder based on an Content-Type header value.
+   *
+   * @param contentType content type header value. If null or not registered the default decoder
+   *     will be returned
+   * @return best Decoder for the given content type header
+   */
+  public Decoder decoder(CharSequence contentType) {
```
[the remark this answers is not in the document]
andyday: yes i misinterpreted.  since there is only one way to get a decoder, i don't think the addition of this preposition clarifies anything.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+/** An Decoder that decodes a JSON ByteBuf to an object. */
+@AllArgsConstructor
+@Accessors(fluent = true)
+public class JsonDecoder implements Decoder {
```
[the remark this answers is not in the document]
andyday: the plan would be for this decoder to use the selected json library through a JsonAdapter or something similar.  this would allow for the json library to be configured without configuring the decoders directly.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
```diff
@@ -0,0 +1,36 @@
...
+import io.netty.buffer.ByteBuf;
+import java.io.IOException;
+
+/** Interface for decoding a request ByteBuf into Object. */
+public interface Decoder {
+  /** Content type this decoder supports. */
+  String contentType();
+
+  /**
+   * Decode a request body from JSON format to an object of designated Class type.
```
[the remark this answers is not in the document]
andyday: fixed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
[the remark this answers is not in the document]
andyday: i was able to use CharSequence by using AsciiSequence.contentEquals.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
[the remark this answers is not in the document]
andyday: this has been addressed with the latest update.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
```diff
@@ -0,0 +1,87 @@
...
+    if (contentType == null) {
+      return defaultDecoder;
+    }
+    // TODO (AD): Handle charset see:
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
+    Decoder decoder = decoders.get(contentType.toString());
```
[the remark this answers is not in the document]
andyday: i've employed my own version of get and getOrDefault in new MediaTypeableCollection base class.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
```diff
@@ -0,0 +1,87 @@
...
+    if (decoder != null) {
+      return decoder;
+    }
+    return defaultDecoder;
+  }
+
+  /** Get Decoders builder. */
+  public static Builder builder() {
```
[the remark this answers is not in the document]
andyday: able to clean this up considerably with lombok generation.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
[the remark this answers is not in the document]
andyday: params are now handled.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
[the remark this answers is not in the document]
andyday: have added #162

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
```diff
@@ -0,0 +1,94 @@
...
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/** Holds a set of Encoder objects each registered to do encoding for a given content type. */
+@Slf4j
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class Encoders {
+  private static final Pattern ACCEPT_SPLIT_PATTERN = Pattern.compile(", *");
+  private final Encoder defaultEncoder;
+  private final ImmutableMap<CharSequence, Encoder> encoders;
```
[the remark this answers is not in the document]
andyday: reworked

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
```diff
@@ -0,0 +1,94 @@
...
+  public Encoder acceptedEncoder(CharSequence accept) {
+    if (accept == null) {
+      return defaultEncoder;
+    }
+
+    // TODO (AD): Consider LRU cache of accepts to encoders.
+    String[] contentTypes = ACCEPT_SPLIT_PATTERN.split(accept);
+    for (String contentType : contentTypes) {
+      // TODO (AD): Handle q-factor weighting see:
+      // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
```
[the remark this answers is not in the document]
andyday: resolved -- todo removed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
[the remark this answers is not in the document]
andyday: fixed.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoders.java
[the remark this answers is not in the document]
andyday: left as decoder.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+  /** Content type this decoder supports. */
+  @Getter private final String contentType;
```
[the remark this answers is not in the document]
andyday: fixed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+   * Decode a ByteBuf body from JSON format to an object of designated Class type.
+   *
+   * @param body current http request
+   * @param clazz target class for decoding
+   * @return object of type clazz
+   */
+  @Override
+  public <T> T decode(ByteBuf body, CharSequence contentType, Class<T> clazz) throws IOException {
+    // TODO (AD): Does this handle encoding properly?
+    // TODO (AD): Close stream?  It does not release bytebuf so no?
```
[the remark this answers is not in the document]
andyday: fixed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+    InputStream input = new ByteBufInputStream(body);
```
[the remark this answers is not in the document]
andyday: fixed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonEncoder.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.OutputStream;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+/** An Encoder that encodes an object to ByteBuf in JSON format. */
+@AllArgsConstructor
+@Accessors(fluent = true)
+public class JsonEncoder implements Encoder {
+  @Getter private final String contentType;
```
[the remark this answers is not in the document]
andyday: fixed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonEncoder.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+/** An Encoder that encodes an object to ByteBuf in JSON format. */
+@AllArgsConstructor
+@Accessors(fluent = true)
+public class JsonEncoder implements Encoder {
```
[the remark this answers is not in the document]
andyday: leaving as is.  see comment above.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoder.java
```diff
@@ -0,0 +1,35 @@
...
+  String contentType();
+
+  /**
+   * Encode an object to ByteBuf.
+   *
+   * @param buf target byte buffer for encoding
+   * @param object object to encode
+   * @return ByteBuf representing encoded object
+   */
+  ByteBuf encode(ByteBuf buf, Object object) throws IOException;
```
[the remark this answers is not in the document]
andyday: done.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.nordstrom.xrpc.encoding;
+
+import io.netty.buffer.ByteBuf;
+import java.io.IOException;
+
+/** Interface for decoding a request ByteBuf into Object. */
+public interface Decoder {
+  /** Content type this decoder supports. */
+  String contentType();
```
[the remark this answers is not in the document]
andyday: great suggestion.  done..

xrpc/src/main/java/com/nordstrom/xrpc/encoding/TextDecodable.java
```diff
@@ -0,0 +1,6 @@
+package com.nordstrom.xrpc.encoding;
+
+/** Interface giving an object to encode itself into a text string. */
```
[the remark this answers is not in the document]
andyday: removed this interface.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/TextDecoder.java
```diff
@@ -0,0 +1,23 @@
...
+ * will fail.
+ */
+@Value
+@Accessors(fluent = true)
+public class TextDecoder implements Decoder {
+  String contentType;
+
+  @Override
+  public <T> T decode(ByteBuf body, CharSequence contentType, Class<T> clazz) throws IOException {
+    throw new NotImplementedException();
```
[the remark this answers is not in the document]
andyday: removed this class for now.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/TextEncoder.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * An Encoder that encodes a response object to Text format. If the Object being encoded implements
+ * TextEncodable, it will be encoded using this interface; otherwise, it will use the Object's
+ * toString method.
+ */
+@AllArgsConstructor
+@Accessors(fluent = true)
+public class TextEncoder implements Encoder {
+  @Getter private String contentType;
```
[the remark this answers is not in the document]
andyday: done.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/TextEncodable.java
```diff
@@ -0,0 +1,6 @@
+package com.nordstrom.xrpc.encoding;
+
+/** Interface giving an object to encode itself into a text string. */
```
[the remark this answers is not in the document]
andyday: Formattable is not quite compatible with what we want to do.  toString is the fallback but there are cases where you don't want to use toString for serialization.  Serializable would work except it would take a bit more work to make it work the way we want.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,82 @@
...
+  private final XrpcRequest request;
+
+  public HttpResponse ok() {
+    return create(HttpResponseStatus.OK, Unpooled.buffer(0), HttpHeaderValues.TEXT_PLAIN);
```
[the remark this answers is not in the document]
andyday: done.  i actually stole this code from the default DefaultHttpRequest constructor.  they propably need to fix as well.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: this was copied over.  will address in next pr.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: will address in next pr

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: how so?  the exceptions still bubble up.  you can still try/catch just like Runtime exception.  in my mind this fixes one of the problems with Java.   checked exceptions SUCK!

xrpc/src/test/java/com/nordstrom/xrpc/EncodingTest.java
andyday: done.

xrpc/src/test/java/com/nordstrom/xrpc/EncodingTest.java
andyday: done.  most of this is done in a split out unit test class

xrpc/src/test/java/com/nordstrom/xrpc/EncodingTest.java
andyday: done.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
andyday: will handle in next pr.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: will consider for next pr.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: will consider for next pr

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: next pr.

xrpc/src/main/resources/com/nordstrom/xrpc/xrpc.conf
andyday: will clarify in next pr.

andyday:
@jkinkead i've addressed most of your comments.  a bit of rework.  anything remaining (unless blocking) i'll try to address in the next pr.

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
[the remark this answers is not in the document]
jkinkead: I'm not sure I agree that it's best-practice to include no-op code, but that's also fine for an example.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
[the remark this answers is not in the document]
jkinkead: SGTM! Is this in the API docs?

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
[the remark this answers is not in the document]
andyday: no-op code?

xrpc/src/main/java/com/nordstrom/xrpc/encoding/TextEncodable.java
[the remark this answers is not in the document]
jkinkead: Yeah, `Serializable` definitely doesn't fit the bill here. We would probably want a special content-type for that.

I think `Formattable` is a logical choice if we consider this to be a debug string, but is a little awkward if we want it to be "encoding".

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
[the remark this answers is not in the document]
jkinkead: I think their use-case is different, because they want people to be able to edit the buffer directly after creation.

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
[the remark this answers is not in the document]
jkinkead: The annotation that doesn't do anything for jackson (I think that's what you were saying here).

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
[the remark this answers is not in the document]
andyday: what API docs?

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
```diff
@@ -0,0 +1,55 @@
...
+import io.netty.handler.codec.http.HttpHeaderValues;
+import io.netty.handler.codec.http.HttpUtil;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import lombok.AllArgsConstructor;
+
+/** An Decoder that decodes a JSON ByteBuf to an object. */
+@AllArgsConstructor
+public class JsonDecoder implements Decoder {
```
jkinkead: I think I've locked access to the previous comment by starting a review...

You said:
> the plan would be for this decoder to use the selected json library through a JsonAdapter or something similar. this would allow for the json library to be configured without configuring the decoders directly.

I think that's a reasonable approach to take, but I also don't like having general-purpose implementations or names for a possible eventuality. It adds abstraction without value.

It's still fine if you want to keep the name `Json`, but you should definitely note in the class-level javadoc that it uses Jackson.

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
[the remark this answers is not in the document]
andyday: the annotation works for Jackson, but is not tied to Jackson.  other libraries use it for decoding as well...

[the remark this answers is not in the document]
andyday: without the annotation, Jackson readValue fails.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/MediaTypeableCollection.java
```diff
@@ -0,0 +1,37 @@
+package com.nordstrom.xrpc.encoding;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import io.netty.util.AsciiString;
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+@Accessors(fluent = true)
+class MediaTypeableCollection<T extends MediaTypeable> {
```
jkinkead: You should probably note performance characteristics (O(N) on number of entries).

You should definitely note why this exists.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonEncoder.java
```diff
@@ -0,0 +1,71 @@
...
+import io.netty.buffer.ByteBufOutputStream;
+import io.netty.handler.codec.http.HttpHeaderValues;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.nio.charset.Charset;
+import lombok.AllArgsConstructor;
+
+/** An Encoder that encodes an object to ByteBuf in JSON format. */
+@AllArgsConstructor
+public class JsonEncoder implements Encoder {
```
jkinkead: Same comment: This should indicate that it's using Jackson, either in its name or in a comment.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonEncoder.java
```diff
@@ -0,0 +1,71 @@
...
+  private static final String[] ALLOWED_CHARSETS = {"UTF-8", "UTF-16", "UTF-32"};
```
jkinkead: `ImmutableSet`?

xrpc/src/main/java/com/nordstrom/xrpc/encoding/TextEncoder.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public ByteBuf encode(ByteBuf buf, CharSequence acceptCharset, Object object) throws IOException {
+    String text =
+        (object instanceof Formattable) ? ((TextEncodable) object).encode() : object.toString();
+    byte[] bytes = text.getBytes(charset(acceptCharset));
+    buf.writeBytes(bytes);
+    return buf;
+  }
+
+  private Charset charset(CharSequence acceptCharset) {
```
jkinkead: Pull into base class, maybe with a TODO?

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
jkinkead: Add TODO?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
jkinkead: You have made it impossible for callers of `body()` to catch the `IOException` that's thrown by it. You've also failed to document that it throws any exceptions.

If you want to wrap in a `RuntimeException` - and document! - that's totally reasonable. `SneakyThrows` is completely unacceptable for expected exceptions - like a malformed request, which will trigger an _uncatchable_ IOException here.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/MediaTypeable.java
```diff
@@ -0,0 +1,6 @@
+package com.nordstrom.xrpc.encoding;
+
+public interface MediaTypeable {
```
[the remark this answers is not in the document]
andyday: object == noun -- or in this case adjectival noun.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: doh.  this annotation has a weird effect.  if you catch Exception and check the type it is IOException but you can't catch (IOException...).   fixed this.

xrpc/src/main/java/com/nordstrom/xrpc/server/Server.java
andyday: done.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: this form of set is not available changed to use forEach

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: fixed

andyday:
@jkinkead i've addressed all latest comments.

andyday:
@jkinkead all future pr comments were resolved also.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoder.java
```diff
@@ -0,0 +1,48 @@
...
+   * Encode an object to ByteBuf.
+   *
+   * @param buf target byte buffer for encoding
+   * @param acceptCharset Accept-Charset header
+   * @param object object to encode
+   * @return ByteBuf representing encoded object
+   */
+  ByteBuf encode(ByteBuf buf, CharSequence acceptCharset, Object object) throws IOException;
+
+  default Charset charset(CharSequence acceptCharset) {
```
jkinkead: Short javadoc?

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Encoders.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Builder
+  private Encoders(CharSequence defaultContentType, @Singular ImmutableList<Encoder> encoders) {
+    super(defaultContentType, encoders);
+  }
+
+  /**
+   * Find an Encoder based on an Accept header value.
+   *
+   * @param accept accept header value.
```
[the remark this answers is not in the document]
jkinkead: Still missing nullability doc...

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonEncoder.java
```diff
@@ -0,0 +1,78 @@
...
+ * An Encoder that encodes an object to ByteBuf in JSON format.
+ *
+ * <p>Currently this encoder uses Jackson ObjectMapper to encode, but eventually will use
+ * configurable JSON encode provider.
+ */
+@AllArgsConstructor
+public class JsonEncoder implements Encoder {
+  public static final Charset DEFAULT_CHARSET = Charset.forName("UTF-8");
+  private static final ImmutableSet<String> ALLOWED_CHARSETS =
+      ImmutableSet.copyOf(new String[] {"UTF-8", "UTF-16", "UTF-32"});
```
jkinkead: Could just be `of("UFT-8", "UTF-16", "UTF-32")`.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/MediaTypeable.java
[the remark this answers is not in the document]
jkinkead: From the style guide: ["Interface names may also be nouns or noun phrases (for example, List), but may sometimes be adjectives or adjective phrases instead (for example, Readable)."](https://google.github.io/styleguide/javaguide.html#s5.2.2-class-names).

`HasMediaType` is an adjective phrase, yes? `MediaTypeable` sounds like something that can be . . . converted into a media type?

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,83 @@
...
+  default <T> HttpResponse internalServerError(T body) throws IOException {
+    return createResponse(HttpResponseStatus.INTERNAL_SERVER_ERROR, body);
+  }
+
+  default <T> HttpResponse createResponse(HttpResponseStatus status, T body) throws IOException {
+    final Encoder encoder =
+        request().connectionContext().encoders().acceptedEncoder(request().acceptHeader());
+    ByteBuf buf = request().byteBuf();
+    return createResponse(
+        status, encoder.encode(buf, request().acceptCharsetHeader(), body), encoder.mediaType());
```
jkinkead: The `Accept` header can also specify a charset, which is actually what I was looking at when I suggested adding charset support. It's not in the HTTP spec, even though [it appears to be supported in the wild](https://stackoverflow.com/questions/7055849/accept-and-accept-charset-which-is-superior); and it's also unclear what priority it should have over the header.

I think a TODO comment is probably appropriate for handling this - especially because charset is mostly interesting for `text/html` anyway, and not for most of what we're doing.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  default HttpResponse createResponse(
+      HttpResponseStatus status,
+      ByteBuf body,
+      CharSequence contentType,
+      Map<String, String> customHeaders) {
+
+    HttpResponse response =
+        body == null || body.readableBytes() <= 0
```
jkinkead: Memoize and conditionally set content headers based on the value?

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -0,0 +1,83 @@
...
+            ? new DefaultHttpResponse(HttpVersion.HTTP_1_1, status)
+            : new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status, body);
+
+    if (customHeaders != null) {
+      customHeaders.forEach((key, value) -> response.headers().set(key, value));
+    }
+
+    response.headers().set(CONTENT_TYPE, contentType);
```
jkinkead: It's weird to send this with no body, but seems harmless...

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -141,37 +139,41 @@ int addData(ByteBuf dataFrame) {
     return h2Data.readableBytes();
   }
 
+  public <T> T body(Class<T> clazz) throws IOException {
```
jkinkead: Javadoc here, especially noting that it consumes the request body (it's not reentrant)?

xrpc/src/test/java/com/nordstrom/xrpc/EncodingFunctionalTest.java
```diff
@@ -0,0 +1,109 @@
...
+  @Test
+  void testJsonDecoding() throws IOException {
+    person = new Person("bob");
+    Response response =
+        client
+            .newCall(
+                new Request.Builder()
+                    .url(endpoint + "/person")
+                    .get()
+                    .header("Accept", "application/json; text/plain")
```
jkinkead: Accept media types should be comma-separated, not semicolon-separated.

xrpc/src/test/java/com/nordstrom/xrpc/EncodingFunctionalTest.java
```diff
@@ -0,0 +1,109 @@
...
+  @Test
+  void testJsonDecodingUtf16() throws IOException {
+    person = new Person("bob");
+    Response response =
+        client
+            .newCall(
+                new Request.Builder()
+                    .url(endpoint + "/person")
+                    .get()
+                    .header("Accept", "application/json;q=1; text/plain")
```
jkinkead: Should be a comma after `q=1`.

xrpc/src/test/java/com/nordstrom/xrpc/EncodingUnitTest.java
```diff
@@ -0,0 +1,139 @@
...
+            .build();
+  }
+
+  @Test
+  void testDefaultEncoder() {
+    assertSame(jsonEncoder, encoders.acceptedEncoder(""));
+  }
+
+  @Test
+  void testDefaultEncoderForNul() {
```
jkinkead: Add missing `l` to method name?

xrpc/src/test/java/com/nordstrom/xrpc/EncodingUnitTest.java
```diff
@@ -0,0 +1,139 @@
...
+  void testEncodeJsonUtf16() throws IOException {
+    assertEquals(
+        Unpooled.copiedBuffer("{\"name\":\"foo\"}", Charset.forName("UTF-16")),
+        jsonEncoder.encode(Unpooled.directBuffer(), "utf-16, foo-bar", new Person("foo")));
+  }
+
+  @Test
+  void testEncodeJsonDefaultCharset() throws IOException {
+    assertEquals(
+        Unpooled.copiedBuffer("{\"name\":\"foo\"}", Charset.forName("UTF-8")),
```
jkinkead: It might be nice to throw a non-ASCII character in this to differentiate UTF-8 from a (bad) ISO-8859-1 encoding being returned.

xrpc/src/test/java/com/nordstrom/xrpc/EncodingUnitTest.java
```diff
@@ -0,0 +1,139 @@
...
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import java.beans.ConstructorProperties;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import lombok.Getter;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class EncodingUnitTest {
```
jkinkead: I'd find this easier to read if the encoder and decoder tests were separate, since they use disjoint fixtures. It's also much more common Java style to have a single unit test class per class under test.

Not a big deal, but it makes this class a lot noisier.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/MediaTypeable.java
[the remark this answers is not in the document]
andyday: agreed MediaTypeable doesn't quite work, but HasMediaType is not an adjective phrase its a verb -> noun fragment.

PR #161, opened 2018-02-28
Fix javadoc warning.

jkinkead:
Just for you, @andyday . :)

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcConnectionContext.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcConnectionContext.java
@@ -42,4 +42,8 @@ public class XrpcConnectionContext {
   @Getter private final CompiledRoutes routes;
 
   @Getter private final ExceptionHandler exceptionHandler;
+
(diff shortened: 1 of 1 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcConnectionContext.java
```diff
@@ -42,4 +42,8 @@
   @Getter private final CompiledRoutes routes;
 
   @Getter private final ExceptionHandler exceptionHandler;
+
+  // This can be generated automatically by lombok, but we declare it here to fix a javadoc warning.
+  // TODO(jkinkead): Remove once we have delombok integrated (issue #160).
+  public static class Builder {}
```
andyday: nice find.  guessing lombok uses this if it finds one already declared...

jkinkead: Yes, it's documented as such. In general, lombok will only create missing things on classes when given class-level annotations. You can definitely break it this way, too. :)

PR #164, opened 2018-03-06
Add proto encoding

andyday:
- ProtoEncoder - supports encoding proto based on Accept header
- ProtoDecoder - supports decoding proto based on Content-Type header
- Modified JsonEncoder to support Json encoding for Proto-based objects
- Modified JsonDecoder to support Json decoding for Proto-based classes
- Modified Decoder to support helper for Class static method invocation.
- Unit Tests.

### DIFF
--- a/xrpc/build.gradle
+++ b/xrpc/build.gradle
@@ -4,6 +4,7 @@ buildscript {
     jcenter()
   }
   dependencies {
+    classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.3'
     classpath 'org.junit.platform:junit-platform-gradle-plugin:1.0.1'
   }
 }
@@ -25,6 +26,7 @@ release {
 }
 
 apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
 apply plugin: 'maven'
 apply plugin: 'maven-publish'
 apply plugin: 'checkstyle'
@@ -45,7 +47,7 @@ tasks.withType(JavaCompile) {
     // the compiler so that the definition works).
     options.compilerArgs += [
         '-XDignore.symbol.file',
-        '-Xlint:auxiliaryclass,cast,classfile,deprecation,dep-ann,divzero,empty,fallthrough,' +
+        '-Xlint:auxiliaryclass,classfile,deprecation,dep-ann,divzero,empty,fallthrough,' +
             'finally,options,overloads,overrides,path,rawtypes,static,try,unchecked,varargs',
         '-parameters',
         '-Werror']
@@ -65,6 +67,8 @@ dependencies {
   compile 'com.squareup.okio:okio:1.13.0'
   compile 'com.typesafe:config:1.3.1'
   compile 'com.xjeffrose:xio:0.13.0'
+  compile 'com.google.protobuf:protobuf-java:3.0.0'
+  compile 'com.google.protobuf:protobuf-java-util:3.0.0'
   compile 'io.dropwizard.metrics:metrics-core:3.2.5'
   compile 'io.dropwizard.metrics:metrics-healthchecks:3.2.5'
   compile 'io.dropwizard.metrics:metrics-json:3.2.5'
@@ -91,6 +95,14 @@ dependencies {
   jmh 'ch.qos.logback:logback-classic:1.1.7'
 }
 
+sourceSets {
+  test {
+    java {
+      srcDirs 'build/generated/source/proto/test/java'
+    }
+  }
+}
+
 // Run with the latest checkstyle version. Required for our checkstyle.xml.
 checkstyle {
   toolVersion '8.3'
@@ -108,6 +120,15 @@ tasks.withType(com.github.spotbugs.SpotBugsTask) {
     html.enabled = true
   }
 }
+
+// Configure protobuf build.
+protobuf {
+  // Configure the protoc executable to use the official distribution.
+  protoc {
+    artifact = 'com.google.protobuf:protoc:3.0.0'
+  }
+}
+
 
 // Add Javadoc + sources jars to artifacts list.
 task javadocJar(type: Jar) {
--- a/xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
@@ -18,6 +18,8 @@ package com.nordstrom.xrpc.encoding;
 
 import io.netty.buffer.ByteBuf;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 
 /** Interface for decoding a request ByteBuf into Object. */
 public interface Decoder extends MediaTypeCodec {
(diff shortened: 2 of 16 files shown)


xrpc/build.gradle
```diff
@@ -45,7 +47,7 @@ tasks.withType(JavaCompile) {
     // the compiler so that the definition works).
     options.compilerArgs += [
         '-XDignore.symbol.file',
-        '-Xlint:auxiliaryclass,cast,classfile,deprecation,dep-ann,divzero,empty,fallthrough,' +
+        '-Xlint:auxiliaryclass,classfile,deprecation,dep-ann,divzero,empty,fallthrough,' +
```
jkinkead: Is `cast` not a valuable linter? It's easy to use the `SuppressWarnings` annotation to get rid of it in a specific place.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
```diff
@@ -48,12 +51,19 @@ public CharSequence mediaType() {
...
+  @SuppressWarnings("unchecked")
   @Override
   public <T> T decode(ByteBuf body, CharSequence contentType, Class<T> clazz) throws IOException {
     try (InputStreamReader reader =
         new InputStreamReader(
             new ByteBufInputStream(body),
             HttpUtil.getCharset(contentType, Charset.forName("UTF-8")))) {
+      if (MessageOrBuilder.class.isAssignableFrom(clazz)) {
+        // Use proto classes to decode
+        Message.Builder builder = invoke(clazz, "newBuilder", new Class<?>[] {}, new Object[] {});
```
jkinkead: Having to go through a reflection layer is slow.

If you replaced the parameter `Class<T> clazz` with `T instance`, you could call `toBuilder()` on the instance if it's of type `Message` (or, technically, `MessageLite`). Message implementations have a `DEFAULT_VALUE` field for just this use.

The use of `clazz` in the jackson serialization can be replaced by `instance.getClass()`.

This should speed up serialization, and avoid the need of the helper for reflecting.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
```diff
@@ -27,14 +30,14 @@
 import lombok.AllArgsConstructor;
 
 /**
- * An Decoder that decodes a JSON ByteBuf to an object.
+ * Decoder that decodes a JSON ByteBuf to an object.
```
jkinkead: I would much prefer having two different JSON encoders / decoders that were swapped in based on a setting or user code. Mixing the two in a single server seems like bad practice.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/ProtoDecoder.java
```diff
@@ -0,0 +1,54 @@
...
+   * @param body current http request
+   * @param clazz target class for decoding
+   * @return object of type clazz
+   */
+  @Override
+  public <T> T decode(ByteBuf body, CharSequence contentType, Class<T> clazz) throws IOException {
+    // TODO (AD): given a Content-Type of application/protobuf; proto=org.some.Message,
+    // we currently ignore the 2nd part, but should at least validate it in the future.
+    try (ByteBufInputStream stream = new ByteBufInputStream(body)) {
+      return invoke(clazz, "parseFrom", new Class<?>[] {InputStream.class}, new Object[] {stream});
```
jkinkead: Similar comment to JSON: If this took an instance, you could call `getParserForType` on it, and avoid the reflection.

xrpc/build.gradle
andyday: this was because of annoying duplicate cast warning in proto generated code...

jkinkead: We should have a TODO to disable this for the protobuf compilation step, then. I don't want the xrpc code to lose a linter warning because of something in a demo project.

Also, isn't the generated code in a different subproject...?

andyday: add todo

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
andyday: fixed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/ProtoDecoder.java
andyday: fixed

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonDecoder.java
andyday: agreed.  tracking with #166

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
```diff
@@ -16,11 +16,18 @@
...
+import com.google.protobuf.MessageLite;
 import io.netty.buffer.ByteBuf;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.concurrent.ConcurrentHashMap;
 
 /** Interface for decoding a request ByteBuf into Object. */
 public interface Decoder extends MediaTypeCodec {
+  /** Cached protobuf MessageLite instances keyed by Class. */
```
jkinkead: Shared global public mutable state is a bit sketchy. I think having this be a private field in an abstract class would be a bit better of a design...

xrpc/src/main/java/com/nordstrom/xrpc/encoding/ProtoDecoder.java
```diff
@@ -0,0 +1,64 @@
...
+   */
+  @Override
+  @SuppressWarnings("unchecked")
+  public <T> T decode(ByteBuf body, CharSequence contentType, Class<T> clazz) throws IOException {
+    // TODO (AD): given a Content-Type of application/protobuf; proto=org.some.Message,
+    // we currently ignore the 2nd part, but should at least validate it in the future.
+
+    Preconditions.checkArgument(
+        MessageLite.class.isAssignableFrom(clazz),
+        String.format("%s does not extend from MessageLite", clazz.getName()));
```
jkinkead: Given the critical-path nature of this code, it's probably worth micro-optimizing this and using a manual throw in an `if`-block to avoid the unneeded formatting call here.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/ProtoEncoder.java
```diff
@@ -0,0 +1,58 @@
...
+   *
+   * @param buf target byte buffer for encoding
+   * @param acceptCharset Accept-Charset header
+   * @param object object to encode
+   * @return ByteBuf representing protobuf formatted bytes
+   */
+  @Override
+  public ByteBuf encode(ByteBuf buf, CharSequence acceptCharset, Object object) throws IOException {
+    if (!(object instanceof MessageLite)) {
+      throw new IllegalArgumentException("object must be instance of MessageLite");
```
jkinkead: This should probably use the same format as the decoder's message.

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
andyday: agreed its not optimal, but making it a private on an abstract class would not help this in any way.  in both cases it would still be mutable and still be shared.  i've added a todo to possibly consider another design which makes it immutable...

andyday:
@jkinkead i've fixed the formatting and added todo

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
jkinkead: The problem now is that it's shared between subclasses, and is mutable by anyone with access to `Decoder` (that is, anyone):
```java
// Clear the cache for some reason.
Decoder.protoDefaultInstances.clear();
```

Private in subclasses fixes these three things (it being public, static, and shared).

xrpc/src/main/java/com/nordstrom/xrpc/encoding/Decoder.java
```diff
@@ -30,4 +39,27 @@
...
   <T> T decode(ByteBuf body, CharSequence contentType, Class<T> clazz) throws IOException;
+
+  /**
+   * Get a protobuf Message instance based on Class. This lazily caches instances as it reflectively
+   * gets them.
+   *
+   * @param clazz proto generated class for which we get the default instance
+   * @return default instance of the given Class
+   */
+  default MessageLite protoDefaultInstance(Class<?> clazz) {
```
jkinkead: This could be static.

andyday:
@jkinkead pulled the proto defaults cache into a separate class which should address the concerns here to a point...

PR #165, opened 2018-03-06
Http2 cors helper

forplayplay:
A CORS helper for HTTP2 requests that plugs into the on headers read method.


xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -62,10 +62,23 @@
...
   Http2Handler(Http2ConnectionEncoder encoder, int maxPayloadBytes) {
     this.encoder = encoder;
     this.maxPayloadBytes = maxPayloadBytes;
 
+    this.corsHandler = new Http2CorsHandler();
+
+    encoder.connection().addListener(this);
+  }
+
+  Http2Handler(Http2ConnectionEncoder encoder, int maxPayloadBytes, Http2CorsHandler corsHandler) {
```
jkinkead: I would either have the two-arg constructor invoke this one directly, or else omit the two-arg constructor entirely.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -102,6 +137,9 @@ private void writeResponse(ChannelHandlerContext ctx, int streamId, HttpResponse
 
     // Convert and validate headers.
     Http2Headers headers = HttpConversionUtil.toHttp2Headers(h1Response, true);
+    if (corsHandler.isCorsSupportEnabled()) {
+      corsHandler.setAccessAllowOriginHeader(headers);
```
jkinkead: This shouldn't be needed (the `writeResponse` call below does this).

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -132,6 +170,10 @@ private void writeResponse(
     headers.setInt(CONTENT_LENGTH, body.readableBytes());
     headers.status(status.codeAsText());
 
+    if (corsHandler.isCorsSupportEnabled()) {
+      corsHandler.setAccessAllowOriginHeader(headers);
```
jkinkead: This shouldn't be needed (the `writeResponse` call below does this).

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -223,6 +265,31 @@ public void onHeadersRead(
...
+    // Handle CORS.
+    if (corsHandler.isCorsSupportEnabled()) {
+      corsHandler.setOrigin(headers);
+      Http2Headers preflightHeaders = corsHandler.preflightHeaders();
+
+      // Friendly short-circuit for OPTIONS requests.
+      if (corsHandler.isPreflightRequest(headers)) {
+        writeResponse(ctx, streamId, preflightHeaders, HttpResponseStatus.OK);
+        ctx.flush();
+        ctx.close();
```
jkinkead: We don't want to close the context here. This disconnects the client, which we don't want. We also don't need a `flush`.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -223,6 +265,31 @@ public void onHeadersRead(
       return;
     }
 
+    // Handle CORS.
+    if (corsHandler.isCorsSupportEnabled()) {
+      corsHandler.setOrigin(headers);
+      Http2Headers preflightHeaders = corsHandler.preflightHeaders();
```
jkinkead: This should happen in an `if`-check.

Something like:
```java
boolean isPreflight = corsHandler.isPreflightRequest(headers);
if (isPreflight || corsHandler.isShortCircuit() && !corsHandler.validateOrigin()) {
  Http2Headers preflightHeaders = ...;
  HttpResponseStatus staus = isPreflight ? HttpResponseStatus.OK : HttpResponseStatus.FORBIDDEN;
  writeResponse(ctx, streamId, preflightHeaders, status);
}
```

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,158 @@
...
+ * <p>This handler can be configured using a {@link CorsConfig}, please refer to this class for
+ * details about the configuration options available.
+ */
+@Slf4j
+public class Http2CorsHandler {
+  private static final String ANY_ORIGIN = "*";
+  private static final String NULL_ORIGIN = "null";
+  private final CorsConfig config;
+
+  @Setter private String origin;
```
jkinkead: I don't like this design, and it's particularly weird for HTTP2. I think this member should go, and methods that need it should take it as a parameter.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -223,6 +265,31 @@ public void onHeadersRead(
       return;
     }
 
+    // Handle CORS.
+    if (corsHandler.isCorsSupportEnabled()) {
+      corsHandler.setOrigin(headers);
```
jkinkead: This is confusing, since it's required for the later `validateOrigin()` call. That's why I suggest changing the API.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -223,6 +265,31 @@ public void onHeadersRead(
...
+      Http2Headers preflightHeaders = corsHandler.preflightHeaders();
+
+      // Friendly short-circuit for OPTIONS requests.
+      if (corsHandler.isPreflightRequest(headers)) {
+        writeResponse(ctx, streamId, preflightHeaders, HttpResponseStatus.OK);
```
jkinkead: You should be using the existing `writeResponse` methods (e.g. provide an empty body explicitly here - `Unpooled.EMPTY_BUFFER`).

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -223,6 +265,31 @@ public void onHeadersRead(
...
+        ctx.close();
+        return;
+      }
+
+      // Friendly short-circuit for CORS requests that don't match allowed origin criteria.
+      // It's important to note that neither this nor the netty cors handler has sophisticated
+      // origin matching so that:
+      // http://some.domain != some.domain and *.domain != some.domain
+      if (corsHandler.isShortCircuit() && !corsHandler.validateOrigin()) {
+        writeResponse(ctx, streamId, preflightHeaders, HttpResponseStatus.FORBIDDEN);
```
jkinkead: This is very weird in HTTP2 because you want to leave the connection open, but discard the stream.

This as-written closes the connection, which we don't want - and if you remove the `close()`, it continues to try to handle the response.

(I just noticed that there's a bug in the rate limiting if the request has a body, also).

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -62,10 +62,23 @@
...
 
+    this.corsHandler = new Http2CorsHandler();
+
+    encoder.connection().addListener(this);
+  }
+
+  Http2Handler(Http2ConnectionEncoder encoder, int maxPayloadBytes, Http2CorsHandler corsHandler) {
+    this.encoder = encoder;
+    this.maxPayloadBytes = maxPayloadBytes;
+    this.corsHandler = corsHandler;
```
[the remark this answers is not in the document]
jkinkead: I think we talked offline about this, but we need to be post-codec in order to read headers.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+
+  protected void setOrigin(Http2Headers headers) {
+    this.origin = headers.get("origin").toString();
+  }
+
+  private static void setOrigin(final Http2Headers headers, final String origin) {
+    headers.set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_ORIGIN, origin);
+  }
+
+  protected boolean isShortCircuit() {
```
andyday: all of these protected s can be prived

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
andyday: there is too much bleed of CORS logic into Http2Handler.  since it would take a bit too much effort now to make the http2 cors handler a ChannelDuplexHandler, it would be good to only expose 2 methods from the cors handler.

1. one that optionally handles preflight and gets the origin from the headers.  this would be called here. 
1. one that writesHeaders it would be called in the writeResponse(...h1Response) method only.

Step 1.  Make a Http2ResponseWriter interface and make the writeResponse with Optional body part of that interface.
Step 2.  Make the following method on Http2CorsHandler

```java
  public boolean handleHeaders(Http2ResponseWriter writer, Http2Headers headers, ChannelHandlerContext ctx, int streamId) {
    if (!isCorsSupportEnabled()) {
      return false;
    }
    setOrigin(headers);

    if (!isPreflightRequest(headers)) {
      return false;
    }
    Http2Headers preflightHeaders = preflightHeaders();

    HttpResponseStatus status = isShortCircuit() && !validOrigin()
      ? HttpResponseStatus.FORBIDDEN
      : HttpResponseStatus.OK;

    preflightHeaders.status(status.codeAsText());

    writer.writeResponse(ctx, streamId, preflightHeaders, Optional.empty());
    return true;
  }
```
Step 3.  call this new method here

```java
    if (corsHandler.handleHeaders(this, headers, ctx, streamId)) {
      // CORS Preflight was handled and response written.
      return;
    }
```

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
andyday: this is the only place needed to write cors headers.  change the name to something like writeHeaders, remove the isCorsSupportEnabled check and add it inside the method.  also in that method you need to handle allow credentials and expose headers as well.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,220 @@
...
+  public boolean handleHeaders(
+      Http2Headers headers,
+      ChannelHandlerContext ctx,
+      int streamId,
+      Http2ResponseWriter responseWriter) {
+    if (!isCorsSupportEnabled()) {
+      return false;
+    }
+
+    setIsPreflight(headers);
```
jkinkead: Per the stateless comment above, these should all inline to local variables, e.g:
```java
// isPreflight instead of setIsPreflight; returns a boolean.
// Most of the methods being called are one-liners, and could / should be inlined.
boolean isPreflight = isPreflight(headers);
```

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -62,9 +62,17 @@
...
   private final int maxPayloadBytes;
 
+  /** Helper for a CORS request. */
+  private final Http2CorsHandler corsHandler;
+
   Http2Handler(Http2ConnectionEncoder encoder, int maxPayloadBytes) {
+    this(encoder, maxPayloadBytes, new Http2CorsHandler());
+  }
+
+  Http2Handler(Http2ConnectionEncoder encoder, int maxPayloadBytes, Http2CorsHandler corsHandler) {
```
jkinkead: You should avoid adding a constructor here.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,220 @@
...
+      }
+      log.debug(
+          "Request origin [{}]] was not among the configured origins [{}]",
+          origin,
+          config.origins());
+    }
+    return false;
+  }
+
+  protected void setOutBoundHeaders(final Http2Headers headers) {
```
jkinkead: Style nit: "outbound" is one word, and should be camel-cased as such.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -103,6 +111,8 @@ private void writeResponse(ChannelHandlerContext ctx, int streamId, HttpResponse
     // Convert and validate headers.
     Http2Headers headers = HttpConversionUtil.toHttp2Headers(h1Response, true);
 
+    corsHandler.setOutBoundHeaders(headers);
```
jkinkead: This needs to go in the `writeResponse` on line 94, as it's the end of the chain (it's the only place bytes are being written on the wire). As written, this will miss calls to the other two `writeResponse` methods.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -241,8 +251,14 @@ public void onHeadersRead(
     // If there's no data expected, call the handler. Else, pass the handler and request through in
     // the context.
     if (endOfStream) {
+      // Handle CORS..
```
jkinkead: Line 168 also needs this logic.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,220 @@
...
+    CorsConfig config = CorsConfigBuilder.forAnyOrigin().disable().build();
+    this.config = config;
+  }
+
+  /** Creates a new instance with the specified {@link CorsConfig}. */
+  public Http2CorsHandler(CorsConfig config) {
+    this.config = checkNotNull(config, "config");
+  }
+
+  public boolean handleHeaders(
```
jkinkead: This should be returning an optional HttpResponse (or a boolean with a second method to fetch the response), instead of being responsible for writing its own output. We're trying to keep I/O localized to the single handler in this model, especially with respect to what's responsible for finalizing requests.

xrpc/src/test/java/com/nordstrom/xrpc/server/Http2HandlerTest.java
```diff
@@ -352,4 +356,135 @@ public void testOnDataRead_payloadTooLarge() {
...
+
+  /** Helper for testing CORS requests. */
+  public CorsConfig corsConfig() {
+    val config = ConfigFactory.load("test.conf").getConfig("xrpc");
+    XConfig xconfig = new XConfig(config);
+    return xconfig.corsConfig();
+  }
+
+  /** Matcher for preflight headers. */
+  ArgumentMatcher<Http2Headers> matchesPreflightHeadersGetRequest(CorsConfig corsConfig) {
```
jkinkead: It would reach much nicer to have a `hasHeader` matcher instead of basing this on the cors config:
```java
hasHeader("access-control-allow-meathods", "GET")
```

We should pair on updating the `verifyResponse` method to allow header specifications as well. It would not be hard.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,220 @@
...
+
+    if ("null".equals(origin) && config.isNullOriginAllowed()) {
+      return true;
+    }
+
+    return config.origins().contains(origin);
+  }
+
+  // outbound
+  protected boolean setAccessAllowOriginHeader(final Http2Headers headers) {
```
andyday: this can be private

xrpc/src/test/java/com/nordstrom/xrpc/server/Http2HandlerTest.java
```diff
@@ -352,4 +356,135 @@ public void testOnDataRead_payloadTooLarge() {
...
+        .add("access-control-request-method", "GET")
+        .path(OK_PATH);
+
+    testHandler.onHeadersRead(mockContext, STREAM_ID, headers, 1, true);
+    assertEquals(1L, requestMeter.getCount());
+    verify(mockEncoder, times(1))
+        .writeHeaders(
+            eq(mockContext),
+            eq(STREAM_ID),
+            argThat(matchesAllowOriginsHeaders(corsConfig)),
```
andyday: better to use argument.capture for these and then assert the captured value .

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
forplayplay: How would we access the headers here?

jkinkead: They're available off of `XrpcRequest`.

jkinkead: (and they're guaranteed to be complete when `endOfStream` is true)

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -241,8 +253,16 @@ public void onHeadersRead(
     // If there's no data expected, call the handler. Else, pass the handler and request through in
     // the context.
     if (endOfStream) {
+      // Handle CORS.
+      Optional<Http2Headers> corsResponseHeaders = corsHandler.inbound(headers);
+      if (corsResponseHeaders.isPresent()) {
+        writeResponse(ctx, streamId, corsResponseHeaders.get(), Optional.of(Unpooled.EMPTY_BUFFER));
```
jkinkead: This is bypassing sending content length and counting the response code; it should use the overload that takes a non-optional `ByteBuf`. Better would be to return an `HttpResponse` from the `inbound` method, and send that directly.

xrpc/src/test/java/com/nordstrom/xrpc/server/Http2HandlerTest.java
```diff
@@ -352,4 +356,135 @@ public void testOnDataRead_payloadTooLarge() {
         Optional.of(Unpooled.wrappedBuffer(XrpcConstants.PAYLOAD_EXCEEDED_RESPONSE)),
         STREAM_ID);
   }
+
+  /** Helper for testing CORS requests. */
+  public CorsConfig corsConfig() {
+    val config = ConfigFactory.load("test.conf").getConfig("xrpc");
```
jkinkead: It's a lot more readable to colocate the test fixtures / configuration with the unit tests, where practical.

I think manually creating the `CorsConfig` in each of the unit tests which use it would be nicer here.

demos/people/src/main/resources/demo.conf
```diff
@@ -12,12 +12,10 @@ xrpc {
...
 
-  # IP White List
-  enable_white_list = true
-  ip_white_list = ["127.0.0.1"]
-
   cors {
     enable = true
     allowed_origins = ["test.domain"]
+    allowed_methods = ["OPTIONS", "GET", "PUT"]
+    short_circuit = true
```
jkinkead: I waffle on whether this should be the recommended default.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.handler.codec.http2.HttpConversionUtil;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Initializes CORS support for this HTTP2 request.
+ *
+ * <ul>
+ *   <li>Request headers are processed by the inbound method, which returns response headers for
```
jkinkead: Update this to reflect new return value? I would also include the third bullet point in here.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+  /** Creates a new instance with the specified {@link CorsConfig}. */
+  public Http2CorsHandler(CorsConfig config) {
+    this.config = checkNotNull(config, "config");
+  }
+
+  /**
+   * Processes inbound HTTP2 CORS requests. Saves the request origin for use in outbound headers.
+   *
+   * @param headers from the request
+   * @return Optional response headers for requests that do not need to continue down the pipeline.
```
jkinkead: Update.

Also, should be lower-cased without a terminal period (the style for sentence fragment javadocs).

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+   */
+  public Optional<HttpResponse> inbound(Http2Headers headers, int streamId) {
+    if (!config.isCorsSupportEnabled()) {
+      return Optional.empty();
+    }
+
+    boolean isPreflight = isPreflight(headers);
+
+    this.requestOrigin = headers.get(HttpHeaderNames.ORIGIN).toString();
```
jkinkead: I would add a comment here along the lines of "Save value for use in outbound header generation."

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+    if (!config.isCorsSupportEnabled()) {
+      return;
+    }
+    if (setAccessAllowOriginHeader(responseHeaders)) {
+      setAllowCredentials(responseHeaders);
+      setExposeHeaders(responseHeaders);
+    }
+  }
+
+  private boolean setAccessAllowOriginHeader(final Http2Headers headers) {
```
jkinkead: A brief doc on this would be good, since it's a critical part of the code.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+      }
+      log.debug(
+          "Request origin [{}]] was not among the configured origins [{}]",
+          requestOrigin,
+          config.origins());
+    }
+    return false;
+  }
+
+  private void echoRequestOrigin(final Http2Headers headers) {
```
jkinkead: Only used once, inline (maybe with comment to replace name).

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+    setOrigin(headers, requestOrigin);
+  }
+
+  private static void setVaryHeader(final Http2Headers headers) {
```
jkinkead: Trivial method used only twice, inline.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+
+  private static void setOrigin(final Http2Headers headers, final String origin) {
+    headers.set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_ORIGIN, origin);
+  }
+
+  private boolean isShortCircuit() {
+    return config.isShortCircuit();
+  }
+
+  private boolean validateOrigin() {
```
jkinkead: Add very short javadoc along the lines of "`Return true if the saved origin is allowed based on the configuration`".

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+
+  private static void setNullOrigin(final Http2Headers headers) {
+    setOrigin(headers, NULL_ORIGIN);
+  }
+
+  private static void setOrigin(final Http2Headers headers, final String origin) {
+    headers.set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_ORIGIN, origin);
+  }
+
+  private boolean isShortCircuit() {
```
jkinkead: Proxy to member, inline instead.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+  }
+
+  private void setExposeHeaders(final Http2Headers headers) {
+    if (!config.exposedHeaders().isEmpty()) {
+      headers.set(HttpHeaderNames.ACCESS_CONTROL_EXPOSE_HEADERS, config.exposedHeaders());
+    }
+  }
+
+  private static void setAnyOrigin(final Http2Headers headers) {
```
jkinkead: Trivial method, inline.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+    setOrigin(headers, ANY_ORIGIN);
+  }
+
+  private static void setNullOrigin(final Http2Headers headers) {
```
jkinkead: Trivial method, inline.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+    setOrigin(headers, NULL_ORIGIN);
+  }
+
+  private static void setOrigin(final Http2Headers headers, final String origin) {
```
jkinkead: This is called in a bunch of places, but it's also trivial. Consider inlining.

If you retain, this should probably be renamed to `setAllowOrigin`.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+    return responseHeaders;
+  }
+
+  private void setAllowMethods(final Http2Headers headers, HttpMethod requestMethod) {
+    if (config.allowedRequestMethods().contains(requestMethod)) {
+      headers.add(HttpHeaderNames.ACCESS_CONTROL_ALLOW_METHODS, requestMethod.toString());
+    }
+  }
+
+  private void setAllowHeaders(final Http2Headers headers) {
```
jkinkead: Trivial method used once, inline.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+    headers.set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_HEADERS, config.allowedRequestHeaders());
+  }
+
+  private void setMaxAge(final Http2Headers headers) {
```
jkinkead: Trivial method used once, inline.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+      setMaxAge(responseHeaders);
+      setExposeHeaders(responseHeaders);
+    }
+    if (!responseHeaders.contains(HttpHeaderNames.CONTENT_LENGTH)) {
+      responseHeaders.set(HttpHeaderNames.CONTENT_LENGTH, HttpHeaderValues.ZERO);
+    }
+    return responseHeaders;
+  }
+
+  private void setAllowMethods(final Http2Headers headers, HttpMethod requestMethod) {
```
jkinkead: Used once, inline.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+
+    boolean isPreflight = isPreflight(headers);
+
+    this.requestOrigin = headers.get(HttpHeaderNames.ORIGIN).toString();
+
+    HttpMethod requestMethod = requestMethod(headers, isPreflight);
+
+    Http2Headers responseHeaders = preflightHeaders(requestMethod);
+    HttpResponseStatus status =
+        isShortCircuit() && !validateOrigin()
```
jkinkead: Given that `validateOrigin` is only called once, from the method setting `requestOrigin`, I would have it take the request origin as an argument instead of reading it from a member variable.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+            ? HttpResponseStatus.FORBIDDEN
+            : HttpResponseStatus.OK;
+    responseHeaders.status(status.codeAsText());
+
+    if (!isPreflight && !status.equals(HttpResponseStatus.FORBIDDEN)) {
```
jkinkead: You should move this short-circuit above the part where you generate response headers. Right now, you're creating response headers even if they're unused.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+      return Optional.empty();
+    }
+
+    HttpResponse response;
+
+    try {
+      response = HttpConversionUtil.toHttpResponse(streamId, responseHeaders, true);
```
jkinkead: It's silly to convert Http2Headers to HttpHeaders only so they can be converted back again. We should probably have a TODO here to fix that.

I suppose that's partly why you were returning headers in the first place. :)

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2Handler.java
```diff
@@ -62,9 +63,20 @@
   /** The maximum total data payload to accept. */
   private final int maxPayloadBytes;
 
+  /** Helper for a CORS request. */
+  private final Http2CorsHandler corsHandler;
+
   Http2Handler(Http2ConnectionEncoder encoder, int maxPayloadBytes) {
```
jkinkead: If this constructor isn't used in our (non-test) codebase, we should remove it.

demos/people/src/main/resources/demo.conf
andyday: me too.  i'd take it out until there's a tangible reason to add it in.

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
andyday: this should be named `validOrigin`

xrpc/src/main/java/com/nordstrom/xrpc/server/Http2CorsHandler.java
```diff
@@ -0,0 +1,230 @@
...
+
+    return config.origins().contains(requestOrigin);
+  }
+
+  /**
+   * Applies appropriate CORS Headers to outbound response headers.
+   *
+   * @param responseHeaders outbound response headers
+   */
+  protected void outbound(final Http2Headers responseHeaders) {
```
andyday: i like the names inbound and outbound

forplayplay: YEA! This actually came from the comment you made about how we need to move this and the Http2Handler to duplex handlers - I wanted to lay some clear groundwork for that but clearly labeling where in the pipeline these take place.

xrpc/src/test/java/com/nordstrom/xrpc/server/Http2HandlerTest.java
```diff
@@ -349,7 +386,65 @@ public void testOnDataRead_payloadTooLarge() {
     assertEquals(0L, requestMeter.getCount());
     verifyResponse(
         HttpResponseStatus.REQUEST_ENTITY_TOO_LARGE,
+        ImmutableMap.of(),
         Optional.of(Unpooled.wrappedBuffer(XrpcConstants.PAYLOAD_EXCEEDED_RESPONSE)),
         STREAM_ID);
   }
+
+  /** Helper for testing CORS requests. */
+  public CorsConfig corsConfig() {
```
jkinkead: I think you can delete this now?

xrpc/src/test/java/com/nordstrom/xrpc/server/Http2HandlerTest.java
andyday: you can also configure the CorsConfig directly without loading the .config.  you're not really testing our config here.

PR #169, opened 2018-03-09
Exception handling

andyday:


xrpc/src/main/java/com/nordstrom/xrpc/server/ExceptionHandler.java
```diff
@@ -14,4 +14,6 @@
    * @return HttpResponse appropriate for this Exception
    */
   HttpResponse handle(XrpcRequest request, Exception exception);
+
+  ExceptionHandler DEFAULT = ResponseFactory::exception;
```
jkinkead: This feels weird as part of the public API of this interface. I also don't like the circular dependency between ResponseFactory and ExceptionHandler.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -14,8 +17,12 @@
 import io.netty.handler.codec.http.HttpVersion;
 import java.io.IOException;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public interface ResponseFactory {
+  Logger log = LoggerFactory.getLogger(ResponseFactory.class);
```
jkinkead: This will log as `ResponseFactory`, which is extremely confusing when a child class uses it (which they probably will, because it's public).

I think I liked it better when there was a `DefaultExceptionHandler` class. :)

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -47,6 +54,37 @@ default HttpResponse ok() {
...
+
+      if (exception instanceof HttpResponseException) {
+        HttpResponseException responseException = (HttpResponseException) exception;
+        return createResponse(
+            HttpResponseStatus.valueOf(responseException.statusCode()), responseException.error());
+      }
+      // TODO (AD): Handle other exceptions that can reasonably be converted to HTTP Responses.
+      // For example IllegalArgumentException could become HTTP 400 Bad Request
+    } catch (Exception e) {
+      log.error("Error Handling Exception:", e);
```
jkinkead: `Error Handling HttpResponseException` instead, to differentiate it from the below log?

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/HttpResponseException.java
```diff
@@ -0,0 +1,68 @@
...
+
+package com.nordstrom.xrpc.exceptions;
+
+import com.nordstrom.xrpc.exceptions.proto.Error;
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+/** Base exception for Exceptions that are meant to convert to HTTP Responses. */
+@Accessors(fluent = true)
+public abstract class HttpResponseException extends RuntimeException {
```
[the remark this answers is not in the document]
andyday: done.

xrpc/src/main/java/com/nordstrom/xrpc/server/ExceptionHandler.java
andyday: removed

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: changed log => errorLog and set this to ExceptionHandler.class .  i left this logic in `ResponseFactory.exception` because trying to keep all response logic here and it also gives users the ability to do `request.exception` in code.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: done

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -62,6 +69,37 @@ default HttpResponse ok() {
     return createResponse(HttpResponseStatus.INTERNAL_SERVER_ERROR, body);
   }
 
+  default HttpResponse exception(Exception exception) {
+    try {
+      // TODO (AD): Use configured response content/type and response specific exceptions to default
+      // to meaningful responses here
+      errorLog.error("Handled Exception:", exception);
```
jkinkead: This should be moved to inside the `instanceof` check, reduced in log level (probably to `info`), and omit the stack trace.

We want the expected case (user is returning like a 404 for a missing resource) to create minimal log noise.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -62,6 +69,37 @@ default HttpResponse ok() {
...
+
+      if (exception instanceof HttpResponseException) {
+        HttpResponseException responseException = (HttpResponseException) exception;
+        return createResponse(
+            HttpResponseStatus.valueOf(responseException.statusCode()), responseException.error());
+      }
+      // TODO (AD): Handle other exceptions that can reasonably be converted to HTTP Responses.
```
jkinkead: I would vote against this feature, as it can mask programming errors.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -62,6 +69,37 @@ default HttpResponse ok() {
...
+      // For example IllegalArgumentException could become HTTP 400 Bad Request
+    } catch (Exception e) {
+      errorLog.error("Error Handling HttpResponseException:", e);
+    }
+    try {
+      // Attempt to return properly encoded response
```
jkinkead: Here is where I'd do an `error` of the exception, with stack trace, since it's unexpected at this point.

xrpc/src/main/proto/error.proto
```diff
@@ -0,0 +1,31 @@
...
+
+syntax = "proto3";
+
+package nordstrom.xrpc;
+
+option java_package = "com.nordstrom.xrpc.exceptions.proto";
+option java_multiple_files = true;
+
+// Error
+message Error {
```
jkinkead: I _might_ rename to like `ErrorResponse`, since it doesn't necessarily mean a server error, and it looks a little odd as a bare `Error`.

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: done

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: removed comment

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
andyday: done.

xrpc/src/main/proto/error.proto
andyday: there's no reason this could not also be used in requests.  i'd prefer to keep it generic for now.

PR #175, opened 2018-03-16
Context Organization Part 1: Rename XrcpConnectionContext to ServerContext

andyday:
This is in prep for having a `ServerContext` which holds server specific contextual data and a `ConnectionContext` which holds connection specific contextual data.  For now I've left the AttributeKey name as CONNECTION_CONTEXT because this will be used for ConnectionContext in Part 2.


xrpc/src/main/java/com/nordstrom/xrpc/server/ServerContext.java
```diff
@@ -31,7 +31,7 @@
 @Builder(builderClassName = "Builder")
 @Accessors(fluent = true)
 // TODO: (AD) Merge with State
-public class XrpcConnectionContext {
+public class ServerContext {
```
jkinkead: It would be good to add javadoc to this while we're cleaning up, explaining what this should hold (state that remains constant for the life of the server, I think).

PR #176, opened 2018-03-17
Context Organization Part 2: Cleanup

andyday:
Since both Http2CorsHandler and Http2Handler are not ChannelHandlers, it will be difficult to make them Sharable and so ConnectionContext is not yet necessary or easy to do...  This PR does some cleanup only...


xrpc/src/main/java/com/nordstrom/xrpc/XrpcConstants.java
```diff
@@ -26,8 +25,6 @@
       AttributeKey.valueOf("XrpcSoftRateLimited");
   public static final AttributeKey<Boolean> XRPC_HARD_RATE_LIMITED =
       AttributeKey.valueOf("XrpcHardRateLimited");
-  public static final AttributeKey<ServerContext> CONNECTION_CONTEXT =
```
jkinkead: 👍

PR #180, opened 2018-04-19
Proto gen for XRPC Server

andyday:
This takes the grpc-java compiler as a starting point and refactors to generate a Service interface adhering to the protobuf definition with methods for each `rpc` and with a default routes() that creates the routes for each `rpc` using `POST /{ServiceName}/{MethodName}` semantics.


andyday:
closing until I git the ci build working.

.travis.yml
```diff
@@ -1,11 +1,17 @@
 language: java
+sudo: required
```
jkinkead: Is this a prereq for `docker`?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# Download dependencies
+RUN gradle resolveDependencies
+
+COPY config/checkstyle config/checkstyle
+
+# Build xrpc-testing
+COPY xrpc-testing/src xrpc-testing/src
+RUN gradle :xrpc-testing:build
+
+RUN ls -al xrpc-testing/
```
jkinkead: Remove.

Dockerfile
```diff
@@ -0,0 +1,46 @@
+# This is used to build all xrpc projects.  It is broken down into parts strategically to take
```
jkinkead: Can you clarify why this is needed, and if it should be used for local development or not?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+COPY demos/people/build.gradle demos/people/
+COPY demos/request-response-pipeline/build.gradle demos/request-response-pipeline/
+
+# Download dependencies
+RUN gradle resolveDependencies
+
+COPY config/checkstyle config/checkstyle
+
+# Build xrpc-testing
+COPY xrpc-testing/src xrpc-testing/src
```
jkinkead: Normally I'd just do a `COPY . .` here, and run a root build target.

.dockerignore
```diff
@@ -0,0 +1,11 @@
+*.md
+LICENSE
+*.iml
+/.gradle/
+/.idea/
+build
```
jkinkead: Does this ignore subdirectory builds as well?

xrpc-proto-gen/java-protoc/Dockerfile
```diff
@@ -0,0 +1,22 @@
+## -*- docker-image-name: "nordstrom/java-protoc" -*-
```
jkinkead: Is this being used at all? Can you comment on what its purpose is here?

xrpc-proto-gen/src/java_plugin/cpp/java_generator.cpp
```diff
@@ -0,0 +1,398 @@
+#include "java_generator.h"
```
jkinkead: So, a ton of this code is copied from Apache-licensed stuff owned by Google. I think the best course of action here is to pull this into a separate git repo (possibly even a fork of `grpc-java`) instead of having to deal with updating the core license & copyright stuff.

.dockerignore
andyday: yes. that is its purpose

.travis.yml
andyday: yes

Dockerfile
andyday: yes.  I need to document it in contributing.md

andyday:
I'm closing this to put it in a separate project.  See #183 for changes that will stay in XRPC.

PR #183, opened 2018-04-24
Add service iface for proto server generation

andyday:
This adds a simple Service interface that can be added to routes.   The proto generated java will implement this interface.


jkinkead:
Awesome, LGTM! Sorry this took so long. :)

PR #186, opened 2018-05-18
br/request added allHeaders and method functions to XrpcRequest

Rivukis:


xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +193,39 @@ public CharSequence header(CharSequence name) {
...
+    if (h1Request != null) {
+      return h1Request.method();
+    } else if (h2Headers != null) {
+      CharSequence rawMethod = h2Headers.method();
+      if (rawMethod != null) {
+        return new HttpMethod(rawMethod.toString());
+      }
+    }
+
+    return null;
```
xjdr: My preference would be to return an Optional<HttpMethod> here vs have the failure more return a null

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +193,39 @@ public CharSequence header(CharSequence name) {
...
+              entry ->
+                  new AbstractMap.SimpleEntry<CharSequence, CharSequence>(
+                      entry.getKey(), entry.getValue()))
+          .collect(Collectors.toList());
+    } else if (h2Headers != null) {
+      List<Map.Entry<CharSequence, CharSequence>> list = new ArrayList<>();
+      h2Headers.iterator().forEachRemaining(list::add);
+      return list;
+    } else {
+      throw new IllegalStateException("Neither HTTP/1 nor HTTP/2 headers set");
```
manimaul: do we really want to throw - why not just return empty map?

jkinkead: This can only be encountered due to an error in xrpc's implementation. It should be impossible, and should trigger a bug report. :)

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +193,39 @@ public CharSequence header(CharSequence name) {
     }
   }
 
+  public List<Map.Entry<CharSequence, CharSequence>> allHeaders() {
```
jkinkead: If you return a `Stream` instead, this becomes simpler & more efficient.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +193,39 @@ public CharSequence header(CharSequence name) {
...
+    } else if (h2Headers != null) {
+      List<Map.Entry<CharSequence, CharSequence>> list = new ArrayList<>();
+      h2Headers.iterator().forEachRemaining(list::add);
+      return list;
+    } else {
+      throw new IllegalStateException("Neither HTTP/1 nor HTTP/2 headers set");
+    }
+  }
+
+  public HttpMethod method() {
```
jkinkead: Add javadoc to public methods, please.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
jkinkead: Is it even possible for the method to be missing in an H2 request at this point? I feel like the codec will likely return an error.

If that's the case, this should throw `IllegalStateException` instead.

Rivukis: I'm fine doing either. Please let me know which approach we want to go with

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: Yeah the issue isn't that there are no headers. It's that we are in a bad state

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +192,39 @@ public CharSequence header(CharSequence name) {
...
 
+  /** Returns a stream of all the HTTP headers. */
+  public Stream<Map.Entry<CharSequence, CharSequence>> allHeaders() throws IllegalStateException {
+    if (h1Request != null) {
+      return h1Request
+          .headers()
+          .entries()
+          .stream()
+          .map(
+              entry ->
```
jkinkead: Is there a reason you can't just return the stream directly, without the `map`?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +192,39 @@ public CharSequence header(CharSequence name) {
     }
   }
 
+  /** Returns a stream of all the HTTP headers. */
+  public Stream<Map.Entry<CharSequence, CharSequence>> allHeaders() throws IllegalStateException {
```
jkinkead: No need for a `throws` declaration for non-checked exceptions; it doesn't do anything.

If there's an unchecked exception that you *expect* to be thrown in some cases, it should go in a javadoc `@throws`.

This particular exception should never happen, and doesn't need documentation.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +192,39 @@ public CharSequence header(CharSequence name) {
...
+    } else if (h2Headers != null) {
+      Iterable<Map.Entry<CharSequence, CharSequence>> iterable = h2Headers::iterator;
+      return StreamSupport.stream(iterable.spliterator(), false);
+    } else {
+      throw new IllegalStateException("Neither HTTP/1 nor HTTP/2 headers set");
+    }
+  }
+
+  /** Returns the HTTP method. */
+  public HttpMethod method() throws IllegalStateException {
```
jkinkead: Same here; remove `throws`.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +192,39 @@ public CharSequence header(CharSequence name) {
...
+    if (h1Request != null) {
+      return h1Request.method();
+    } else if (h2Headers != null) {
+      CharSequence rawMethod = h2Headers.method();
+      if (rawMethod != null) {
+        return new HttpMethod(rawMethod.toString());
+      }
+    }
+
+    throw new IllegalStateException("Neither HTTP/1 nor HTTP/2 method set");
```
jkinkead: Did you confirm that `method()` will never return null?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: Everything else in this class returns CharSequence so I figured this should as well. The map is turning the Strings into CharSequence. If that's not necessary then I'll remove it

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: Nope, I went with Chris's advise that we shouldn't end up in this state. If that's not the case or I misinterpreted what he said please let me know

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: fixed

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: fixed

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: Also, if I change it to String then I'll have to make the h2 headers instead. So either way I think I'll have to do a map.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: Nvm I mis-read the earlier comment. I'll change this to return an optional

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
jkinkead: I didn't realize that these were different types. I fail to understand why `String`, which is a `CharSequence`, can't just fill here without rebuilding objects - but I suppose this is fine.

If you really wanted to do this without rewrapping, you could have this method instead take a `BiConsumer<CharSequence, CharSequence>`, which you could pass all header values to.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: I changed it to return an optional

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
jkinkead: If you're being thorough, you should return an `Optional` from the `else` block above, and leave the `IllegalState` throw here.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -188,6 +193,39 @@ public CharSequence header(CharSequence name) {
...
+
+    throw new IllegalStateException("Neither HTTP/1 nor HTTP/2 headers set");
+  }
+
+  /** Returns the HTTP method. */
+  public Optional<HttpMethod> method() {
+    if (h1Request != null) {
+      return Optional.of(h1Request.method());
+    } else if (h2Headers != null) {
+      CharSequence rawMethod = h2Headers.method();
```
jkinkead: More clearly, this should be:

```java
  return Optional.ofNullable(h2Headers.method()).map(rawMethod -> new HttpMethod(rawMethod.toString());
} else {
  throw new IllegalStateException("...");
}
```

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
Rivukis: I'd rather not have to have the consumer pass in a parameter just to solve an internal issue. Thoughts?

jkinkead: I don't think we care that much about the extra garbage creation that's happening here. Leaving it as-is is fine (but a comment at the top along the lines of `// Convert to CharSequence.` would be helpful).

PR #194, opened 2018-06-02
change-to-incorporate-custom-exception

shine17:
* Modified the error method to return an generic object
* Added a sample route to demonstrate custom exception response payload
Now users can send custom error payload to clients
eg : custom error payload(users can define it)
```
{"businessReason":"Some business reason","businessStatusCode":"4.2.3000"}
```

Earlier xrpc  error payload properties will always be like below (errorCode,message)
```
{"errorCode":"Badrequest","message":"Invalid input"}
```


xrpc/src/main/java/com/nordstrom/xrpc/exceptions/HttpResponseException.java
```diff
@@ -67,12 +67,11 @@ public HttpResponseException(int statusCode, String errorCode, String message, T
...
   /**
-   * Get the error response for this exception. This is used for response encoding. Error is a proto
-   * generated class that can encode to proto and json.
+   * Get the error response for this exception. This is used for response encoding.
    *
-   * @return Error proto generated object based on this exception.
+   * @return Error object based on this exception.
    */
-  public Error error() {
+  public Object error() {
```
andyday: I would do the following here:

```java
public class HttpResponseException extends RuntimeException {
    @Getter private final int statusCode;
    @Getter private final Object error;

  public HttpResponseException(int statusCode, Object error, String message) {
    super(message);
    Preconditions.checkArgument(
         statusCode >= 200 && statusCode <= 527, "statusCode should be >= 200 && <= 527");
     Preconditions.checkArgument(error == null, "error should be non-null");
     this.statusCode = statusCode;
     this.error = error;
}

...

class BadRequestException extends HttpResponseException {
   public BadRequestException(String message) {
      super(STATUS_CODE, 
         Error.newBuilder().setErrorCode(errorCode).setMessage(message).build(),
         message);
    }
}
```

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
```diff
@@ -38,6 +38,19 @@ public PeopleRoutes() {
     get("/people", getPeople);
     post("/people", postPerson);
     get("/people/{person}", getPerson);
+
+    Handler customException =
```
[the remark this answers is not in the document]
andyday: I would also do the new demo in a new pr and add a unit test to the exceptions package.

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/HttpResponseException.java
shine17: Address in latest commit. Included error object in constructor and removed the mutation.
Also added null check for jsonencoder and proto encoder

demos/people/src/main/java/com/nordstrom/xrpc/demos/people/PeopleRoutes.java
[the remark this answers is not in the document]
shine17: moved the exception demos to a new project

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/BadRequestException.java
```diff
@@ -19,24 +19,25 @@
...
    * Construct BadResponseException with default error code.
    *
    * @param message Error message
+   * @param error Error object
    */
-  public BadRequestException(String message) {
-    super(STATUS_CODE, ERROR_CODE, message);
+  public BadRequestException(String message, Object error) {
+    super(STATUS_CODE, error, message);
   }
```
andyday: this should be

```java
public BadRequestException(String message) {
  super(Error.builder().message(message).errorCode(ERROR_CODE).build())
}
```

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/BadRequestException.java
```diff
@@ -19,24 +19,25 @@
 /** Exception to be converted to an HTTP 400 Bad Request response. */
 public class BadRequestException extends HttpResponseException {
   private static final int STATUS_CODE = 400;
-  private static final String ERROR_CODE = "BadRequest";
```
andyday: This should be reverted.  See below.

settings.gradle
```diff
@@ -2,6 +2,7 @@ include 'xrpc'
 include 'xrpc-testing'
 include 'demos:dino'
 include 'demos:people'
+include 'demos:exceptions'
```
andyday: Please move the exception demo to a separate PR once this PR is done...

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/BadRequestException.java
andyday: This applies to all of the specialized exceptions.

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/HttpResponseException.java
```diff
@@ -28,68 +26,46 @@
...
-
-  @Override
-  public String toString() {
-    StringBuilder buffer =
-        new StringBuilder()
-            .append(getClass().getName())
-            .append(": [")
-            .append(statusCode())
-            .append("] ")
-            .append(errorCode());
```
andyday: We should have an intermediate class that all of the specialized exception classes extend from.

```java
@Accessors(fluent = true)
public class DefaultHttpResponseException extends HttpResponseException {
    @Getter private final string errorCode;

    public DefaultHttpResponseException(int statusCode, String errorCode, String message) {
        super(statusCode, message, Error.builder().message(message).errorCode(errorCode).build());
        this.errorCode = errorCode;
    }

    public DefaultHttpResponseException(int statusCode, String errorCode, String message, Throwable cause) 
        super(statusCode, message, Error.builder().message(message).errorCode(errorCode).build(), cause);
        this.errorCode = errorCode;
   }

  @Override
  public String toString() {
     StringBuilder buffer =
        new StringBuilder()
            .append(getClass().getName())
            .append(": [")
            .append(statusCode())
            .append("] ")
            .append(errorCode());
    String message = getLocalizedMessage();
    if (!Strings.isNullOrEmpty(message)) {
      buffer.append(": ").append(message);
    }
    return buffer.toString();
  }
}
```
and then implement it with the specialized exceptions like this

```java
public class BadRequestException extends DefaultHttpResponseException {
  private static final int STATUS_CODE = 400;
  private static final String ERROR_CODE = "BadRequest";

  public BadRequestException(String message) {
    super(STATUS_CODE, ERROR_CODE, message);
  }

  public BadRequestException(String message, Throwable cause) {
    super(STATUS_CODE, ERROR_CODE, message, cause);
  }
}
```

shine17: Addressed in latest commit

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/BadRequestException.java
shine17: Addressed in latest commit

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/BadRequestException.java
shine17: Addressed in latest commit

settings.gradle
shine17: Addressed in latest commit

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
```diff
@@ -88,7 +88,7 @@ default HttpResponse exception(Exception exception) {
       errorLog.error("Handled Exception:", exception);
       return createResponse(
           HttpResponseStatus.INTERNAL_SERVER_ERROR,
-          new InternalServerErrorException("Internal Server Error").error());
+          new InternalServerErrorException("Internal Server Error", null).error());
```
andyday: should remove the `null`

xrpc/src/test/java/com/nordstrom/xrpc/exceptions/ErrorResponse.java
```diff
@@ -0,0 +1,13 @@
+package com.nordstrom.xrpc.exceptions;
```
andyday: This class does not belong here.

xrpc/src/test/java/com/nordstrom/xrpc/exceptions/ErrorResponse.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@Builder
+public class ErrorResponse {
```
andyday: This should be a `@Value` immutable object

xrpc/src/test/java/com/nordstrom/xrpc/exceptions/ExceptionTest.java
```diff
@@ -151,10 +153,40 @@ void testUnhandled() throws IOException {
...
+  private static String jsonString(Object object) {
+    ObjectMapper objectMapper = new ObjectMapper();
+    String jsonString = null;
+    try {
+      jsonString = objectMapper.writeValueAsString(object);
+    } catch (JsonProcessingException e) {
+      e.printStackTrace();
+    }
+    return jsonString;
+  }
```
andyday: this can be simplified to:

```java
private static String jsonString(Object object) {
   return new ObjectMapper().writeValueAsString(object);
}
```

Catching the Json exception and printing is not necessary as Junit will do a proper job of reporting the exception.  Also declaring `jsonString` outside of the try/catch is not necessary.  The `objectMapper` call can be returned directly.

xrpc/src/test/java/com/nordstrom/xrpc/exceptions/MyCustomException.java
```diff
@@ -0,0 +1,10 @@
+package com.nordstrom.xrpc.exceptions;
+
+public class MyCustomException extends HttpResponseException {
```
andyday: The idea with this customization is to have the custom exception hide the details of the ErrorResponse.  So rather than taking an `Object myError` param, it would take `String businessReason, String businessStatusCode` and construction the ErrorResponse internally.

shine17: Done.Addressed in latest commit

xrpc/src/main/java/com/nordstrom/xrpc/server/ResponseFactory.java
shine17: Done.Addressed in latest commit

xrpc/src/test/java/com/nordstrom/xrpc/exceptions/ExceptionTest.java
shine17: Done.Addressed in latest commit

xrpc/src/test/java/com/nordstrom/xrpc/exceptions/ErrorResponse.java
shine17: Done.Addressed in latest commit

andyday:
👍 LGTM!

xrpc/src/main/java/com/nordstrom/xrpc/encoding/JsonEncoder.java
```diff
@@ -66,7 +66,9 @@ public ByteBuf encode(ByteBuf buf, CharSequence acceptCharset, Object object) th
         writer.write(json);
       } else {
         // Encode POJO
-        mapper.writeValue(writer, object);
+        if (object != null) {
```
jkinkead: This is counter to how we serialize other objects with Jackson, I believe. The default is to serialize nulls as JSON nulls, and I don't think we change this setting. Is there a reason you did this?

PR #195, opened 2018-06-25
Add request context id property in xrpc request object

shine17:
* This property will generate a GUID for each request
* This value can be used while logging exceptions and request/response logs
* This will help in debugging in case 4XX and 5Xx errors
* Also this value can be used as custom header to upstream calls for tracing and debugging purpose

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
@@ -32,6 +32,7 @@ import java.nio.charset.Charset;
 import java.util.AbstractMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import lombok.AccessLevel;
@@ -55,6 +56,13 @@ public class XrpcRequest implements ResponseFactory {
   @Getter private final ByteBufAllocator alloc;
 
   @Getter private final ServerContext connectionContext;
+
+  /**
+   * Unique GUID assigned to each incoming request. This value can be used as part of
+   * request/response logging, exception logging, custom http header to upstream calls for tracing
+   * etc
+   */
+  @Getter private final String requestContextId;
 
   /** The variables captured from the route path. */
   private final Map<String, String> groups;
@@ -80,6 +88,7 @@ public class XrpcRequest implements ResponseFactory {
     this.alloc = channel.alloc();
     this.eventLoop = channel.eventLoop();
     this.h2Data = null;
+    this.requestContextId = UUID.randomUUID().toString();
   }
 
   public XrpcRequest(
@@ -95,6 +104,7 @@ public class XrpcRequest implements ResponseFactory {
     this.alloc = channel.alloc();
     this.eventLoop = channel.eventLoop();
     this.h2Data = alloc.compositeBuffer();
+    this.requestContextId = UUID.randomUUID().toString();
   }
 
   public HttpQuery query() {


xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -80,6 +88,7 @@ public XrpcRequest(
     this.alloc = channel.alloc();
     this.eventLoop = channel.eventLoop();
     this.h2Data = null;
+    this.requestContextId = UUID.randomUUID().toString();
```
andyday: I would change this to `id` since this is the Request.  `requestContextId` is a bit redundant...

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -56,6 +57,13 @@
 
   @Getter private final ServerContext connectionContext;
 
+  /**
+   * Unique GUID assigned to each incoming request. This value can be used as part of
```
andyday: UUID not GUID...

andyday: actually just call it a unique ID.  Also if this will be used for log correlation, this should use the SlugId method in the standards to generate the ID...

shine17:
current branch 
xrpc-pr-branch

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
shine17: Addressed in latest commit

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
shine17: Addressed in latest commit. added the random generator as per standards

xrpc/src/main/java/com/nordstrom/xrpc/utils/RandomIdGenerator.java
```diff
@@ -0,0 +1,23 @@
+package com.nordstrom.xrpc.utils;
+
+import java.nio.ByteBuffer;
+import java.util.Base64;
+import java.util.UUID;
+
+public class RandomIdGenerator {
```
jkinkead: This is very specifically a request ID generator, and should be named appropriately (`RequestIdGenerator`).

xrpc/src/main/java/com/nordstrom/xrpc/utils/RandomIdGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+  public static String generate() {
```
jkinkead: Suggested rewrite, using Google's libraries, which Do the Right Thing:

```java
import com.google.common.io.BaseEncoding;

/** Singleton encoder. */
private static final encoder = BaseEncoding.base64Url().omitPadding();

public static String generate() {
  UUID uuid = UUID.randomUUID();
  ByteBuffer buffer = ByteBuffer.allocate(16);
  buffer.putLong(uuid.getMostSignificantBits());
  buffer.putLong(uuid.getLeastSignificantBits());
  byte[] bytes = buffer.array();
  // NOTE: This fixes a bug in your code.
  bytes[0] = bytes[0] & 0x7f;
  return encoder.encode(bytes);
}
```

andyday: He's using the prescribed logic to generate a 10 char shorter slugid in keeping with logging standards.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -80,6 +87,7 @@ public XrpcRequest(
     this.alloc = channel.alloc();
     this.eventLoop = channel.eventLoop();
     this.h2Data = null;
+    this.id = RandomIdGenerator.generate();
```
andyday: Ideally this should only be generated if there is not the designated header value.  Default is `X-Request-Id` but should be configurable.  See logging standards.

xrpc/src/main/java/com/nordstrom/xrpc/utils/RandomIdGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+    UUID uuid = UUID.randomUUID();
+    ByteBuffer bytes = ByteBuffer.wrap(new byte[16]);
+    bytes.putLong(uuid.getMostSignificantBits());
+    bytes.putLong(uuid.getLeastSignificantBits());
+    // unset most significant bit of first byte
+    // to ensure id starts with [A-Za-f]
+    byte[] byteArr = bytes.array();
+    byteArr[0] = 0x7f;
+    return Base64.getEncoder()
```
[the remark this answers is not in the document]
shine17: Addressed in latest commit . Updated the code as per the standards.

xrpc/src/main/java/com/nordstrom/xrpc/utils/RandomIdGenerator.java
shine17: Addressed in latest commit .

xrpc/src/main/java/com/nordstrom/xrpc/utils/RandomIdGenerator.java
shine17: Addressed in latest commit . Updated the code as per the standards. Used google libraries.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
shine17: I guess the header `X-Request-Id` is something which is defined by our standards. Not sure if that is applicable for a github project. Again in our case, we have different internal teams which uses the header with different names like `TraceContext` , `CorrelationId` etc. So it's really difficult to code for a particular request header. What I intended to do with this is to get a correlation of xrpc error logs with xrpc request/response logs , so that we can narrow down exception logs related to 4XX or 5XX in elastic search-kibana/splunk or any log parsing tool

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -56,6 +57,12 @@
 
   @Getter private final ServerContext connectionContext;
 
+  /**
+   * Unique Id assigned to each incoming request. This value can be used as part of request/response
+   * logging, exception logging, custom http header to upstream calls for tracing etc
+   */
+  @Getter private final String id;
```
jkinkead: Rather than being a final field generated automatically, this should be a lazily-initialized field that's read from a configurable header field, per @andyday 's comment. Having it auto-generate if it's missing is fine, but should be disableable (we don't want people to be unable to tell if a request came in with a header or not).

shine17: Addressed in latest commit. The slug will be set only when request response logging is enabled. Also will check the request header for `x-request-id` before generating the slug. 
The sad part is I have to do a lot of null checks to cover all the scenerio where server context will be null or headers will be null

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
shine17: Addressed in latest commit. The slug will be set only when request response logging is enabled. Also will check the request header for `x-request-id` before generating the slug. 
The sad part is I have to do a lot of null checks to cover all the scenerio where server context will be null or headers will be null

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -236,4 +249,20 @@ public CharSequence contentTypeHeader() {
   public CharSequence acceptCharsetHeader() {
     return header(HttpHeaderNames.ACCEPT_CHARSET);
   }
+
+  private Optional<String> getRequestId() {
+    if (this.connectionContext != null
+        && this.connectionContext.config() != null
+        && this.connectionContext.config().requestResponseLogging()) {
+      if (this.h1Request != null
```
jkinkead: Use the existing helper:

```java
String header = header(X_REQUEST_ID);
```

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -236,4 +249,20 @@ public CharSequence contentTypeHeader() {
...
+          && this.h1Request.headers() != null
+          && Optional.ofNullable(h1Request.headers().get(xrequestId)).isPresent()) {
+        return Optional.ofNullable(h1Request.headers().get(xrequestId));
+      } else if (this.h2Headers != null && this.h2Headers.get(xrequestId) != null) {
+        return Optional.ofNullable(this.h2Headers.get(xrequestId).toString());
+      }
+      return Optional.ofNullable(RequestIdGenerator.generate());
```
jkinkead: This isn't nullable; use `of` instead.

You can also chain with my previous example:
```java
return Optional.of(header == null ? RequestIdGenerator.generate() : header);
```

xrpc/src/main/resources/com/nordstrom/xrpc/xrpc.conf
```diff
@@ -70,6 +70,8 @@ slf4j_reporter_polling_rate = 30
 
 jmx_reporter = true
 
+enable_request_response_logging = false
```
jkinkead: This should be renamed; it enables a unique ID, and should be named as such. Maybe `enable_request_id_generation`?

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -80,6 +91,7 @@ public XrpcRequest(
     this.alloc = channel.alloc();
     this.eventLoop = channel.eventLoop();
     this.h2Data = null;
+    this.id = getRequestId();
```
jkinkead: You shouldn't invoke member functions from a constructor, *especially* if you're using the return value to populate another member. It can leak uninitialized objects, and worse, creates an ordering dependency between the members of your class.

This `getRequestId` should be rewritten to be a static function.

xrpc/src/main/java/com/nordstrom/xrpc/server/ServerContext.java
```diff
@@ -50,6 +51,8 @@
 
   @Getter private final Decoders decoders;
 
+  @Getter private final XConfig config;
```
andyday: We've been avoiding adding XConfig to ServerContext.  You should instead pass the appropriate logging config here.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
```diff
@@ -44,6 +44,9 @@
 @Slf4j
 @Accessors(fluent = true)
 public class XrpcRequest implements ResponseFactory {
+  /** The x-request-id header should be in lower case for http2. */
+  private static final String xrequestId = "x-request-id";
```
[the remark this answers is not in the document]
andyday: this header name should be configurable.  see above comment about config format.  also this should not be called xrequestid.  I would call it requestId instead -- again see above comment.

xrpc/src/main/java/com/nordstrom/xrpc/server/XrpcRequest.java
andyday: Rather than adding this `getRequestId` to `XrpcRequest` and using it in this fashion, perform the request Id lookup logic prior to constructing the request in the calling handlers and pass it in as an Optional<String> to the constructors.

xrpc/src/main/java/com/nordstrom/xrpc/XConfig.java
```diff
@@ -159,6 +160,7 @@ public XConfig(Config configOverrides) {
     corsConfig = buildCorsConfig(config.getConfig("cors"));
 
     populateClientOverrideList(config.getObjectList("req_per_second_override"));
+    requestResponseLogging = config.getBoolean("enable_request_response_logging");
```
andyday: Given the fact that we also need the request id header to be configurable I would change the config to the following

```
request_logging {
   enable = false
   request_id_header = "x-request-id"
}
```

and then create 

```java
@Value
class RequestLoggingContext {
   private final boolean enable;
   private final String requestIdHeader;
}
```

and this pass this object into ServerContext at creation.  This properly decouples internal context from external config.

PR #196, opened 2018-06-26
Don't allow null response bodies to be serialized.

jkinkead:
This fixes a behavior change introduced in #194. There are response builder helpers that don't take a body, and should be used for an empty response body. A null body probably indicates a programming error.

Clean up a exception construction slightly, and add back in preconditions checks which were removed. This forces the user to provide an error message, which is what we want (information-free errors are not great).

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/exceptions/DefaultHttpResponseException.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/exceptions/DefaultHttpResponseException.java
@@ -1,29 +1,21 @@
 package com.nordstrom.xrpc.exceptions;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.nordstrom.xrpc.exceptions.proto.Error;
-import lombok.Getter;
-import lombok.experimental.Accessors;
 
-@Accessors(fluent = true)
+/** Base exception for responses using a simple string error code in the response. */
 public class DefaultHttpResponseException extends HttpResponseException {
-  @Getter private final String errorCode;
+  private final String errorCode;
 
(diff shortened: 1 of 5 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/exceptions/DefaultHttpResponseException.java
```diff
@@ -1,29 +1,21 @@
...
 import com.google.common.base.Strings;
 import com.nordstrom.xrpc.exceptions.proto.Error;
-import lombok.Getter;
-import lombok.experimental.Accessors;
 
-@Accessors(fluent = true)
+/** Base exception for responses using a simple string error code in the response. */
 public class DefaultHttpResponseException extends HttpResponseException {
-  @Getter private final String errorCode;
+  private final String errorCode;
```
andyday: exposing this with a fluent getter could be useful in certain scenarios, especially if they are implementing their own exception handler.  I definitely see no need to remove it.

xrpc/src/main/java/com/nordstrom/xrpc/exceptions/DefaultHttpResponseException.java
```diff
@@ -35,12 +27,23 @@ public String toString() {
...
     }
     return buffer.toString();
   }
+
+  /**
+   * Builds an error message from the given fields.
+   *
+   * @throws IllegalArgumentException if errorCode is empty or null
+   */
+  private static Error buildError(String errorCode, String message) {
```
andyday: this cleans things up a bit.  thanks.

andyday:
👍 LGTM!

PR #197, opened 2018-07-19
Allow CRLF in raw certificate string

mindyor:
Cleans up the raw certificate string by replacing CRLFs with LFs.
This makes parsing the string a bit more permissive.

We ran into this issue while trying to manually chain certs in a file.

### DIFF
--- a/xrpc/src/main/java/com/nordstrom/xrpc/server/tls/Tls.java
+++ b/xrpc/src/main/java/com/nordstrom/xrpc/server/tls/Tls.java
@@ -16,7 +16,8 @@
 
 package com.nordstrom.xrpc.server.tls;
 
-import com.nordstrom.xrpc.XrpcConstants;
+import static com.nordstrom.xrpc.server.tls.X509CertificateGenerator.parseX509Certificates;
+
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.ChannelHandler;
 import io.netty.handler.codec.http2.Http2SecurityUtil;
@@ -31,7 +32,6 @@ import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.ssl.SslProvider;
 import io.netty.handler.ssl.SupportedCipherSuiteFilter;
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.security.InvalidKeyException;
 import java.security.KeyStore;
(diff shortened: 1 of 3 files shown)


xrpc/src/main/java/com/nordstrom/xrpc/server/tls/Tls.java
```diff
@@ -16,7 +16,8 @@
 
 package com.nordstrom.xrpc.server.tls;
 
-import com.nordstrom.xrpc.XrpcConstants;
+import static com.nordstrom.xrpc.server.tls.X509CertificateGenerator.parseX509Certificates;
```
jkinkead: Can you remove the static import and import the `X509CertificateGenerator` class instead? You can invoke the method there.

This helps with readability, since it's clear at call point where the symbol is coming from.

Alternately, just leave the helper code in this class?

mindyor: Whoops, fair point on readability and avoiding import duplication.

Waffled on it a bit and ultimately left the method in X509CertificateGenerator. If that seems off with the intention of the class, happy to move it back to being a helper method in Tls.

xrpc/src/main/java/com/nordstrom/xrpc/server/tls/X509CertificateGenerator.java
```diff
@@ -174,6 +177,32 @@ public static X509Certificate generate(String keyPath, String certPath) {
...
+      certList.add(x509Certificate);
+    }
+
+    chain = new java.security.cert.X509Certificate[certList.size()];
+
+    for (int i = 0; i < certList.size(); i++) {
+      chain[i] = certList.get(i);
+    }
+    return chain;
+  }
```
andyday: This method can be simplified to

```java
  public static java.security.cert.X509Certificate[] parseX509Certificates(String rawCertString)
      throws CertificateException {
    rawCertString = rawCertString.replace("\r\n", "\n");
    CertificateFactory cf = CertificateFactory.getInstance("X.509");
    val rawCerts = rawCertString.split("-----END CERTIFICATE-----\n");
    val certs = new java.security.cert.X509Certificate[rawCerts.length];
    for (int i = 0; i < rawCerts.length; i++) {
      certs[i] =
          (java.security.cert.X509Certificate)
              cf.generateCertificate(
                  new ByteArrayInputStream(
                      (rawCerts[i] + "-----END CERTIFICATE-----\n").getBytes(DEFAULT_CHARSET)));
    }
    return certs;
  }
```

andyday:
@jkinkead this is superseded by #207 no?

PR #199, opened 2018-08-06
Mutual auth. Take two.

Farhie:
• Adds tests around default config values for both `XRPC` and `TLS`
• Add tests around server initialisation (are the various admin endpoints bound by default)
• Generates self-signed certificates (*.nordstrom.com) if `x509CertPath` or `privateKeyPath` are not supplied
• Default cyphersuite now lines up with engineering standards.

**Note on XIO**: We ended up overriding XIO `TlsConfig` as it had a hard requirement for a certificate to be present in a file written to disk. We didn't want this behaviour. We can put a PR or open an issue with the XIO repo to allow cert / private key to be env vars or paths to files.


config/checkstyle/suppressions.xml
```diff
@@ -5,4 +5,5 @@
   "http://checkstyle.sourceforge.net/dtds/suppressions_1_1.dtd">
 <suppressions>
   <suppress files="[\\/]build[\\/]generated[\\/]" checks="[a-zA-Z0-9]*"/>
+  <suppress checks="PackageName" files="TlsConfig.java"/>
```
jkinkead: Can you fix the style instead of disabling?

xrpc/src/main/java/com/xjeffrose/xio/SSL/TlsConfig.java
```diff
@@ -0,0 +1,251 @@
+package com.xjeffrose.xio.SSL;
```
jkinkead: Why are you copying in a full file instead of updating the xio dep?

xrpc/src/main/java/com/xjeffrose/xio/SSL/TlsConfig.java
```diff
@@ -0,0 +1,251 @@
...
+import java.security.cert.X509Certificate;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.ArrayList;
+import java.util.List;
+import javax.xml.bind.DatatypeConverter;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class TlsConfig {
```
jkinkead: Is there a way you could extend the `xio` class instead of copying it in whole?

Farhie: Thanks for your feedback. The problem comes from the fact that the constructor for TlsConfig under `xio` requires that `x509CertPath` and `privateKeyPath` are a path to the appropriate files. If it's not supplied, TlsConfig blows up. 
```java
public TlsConfig(Config config) {
  ...
  certificate = parseX509CertificateFromPem(readPathFromKey("x509CertPath", config));
  privateKey = parsePkcs8FormattedPrivateKeyFromPem(readPathFromKey("privateKeyPath", config));
  ...
}
```
If it's not supplied, we want to generate self-signed certs. We could have a method that:
• Checks if `x509CertPath`, `privateKeyPath` are specified
• If not, generate the cert and private key
• Write the cert and private key to disk
• Override the `x509CertPath` and `privateKeyPath` within the config block with the path of the generated cert and key
• TlsConfig will then parse those files back into the object they were before being written to disk

This feels like somewhat of an awkward workflow and rather convoluted. 

As alluded to in the note on the PR, I think the actual fix for this is a PR into XIO, to generate self-signed certs if they aren't specified.

xrpc/src/main/java/com/xjeffrose/xio/SSL/TlsConfig.java
Farhie: Thanks for the feedback.

As mentioned in the note on the opening of the PR, that's what the proper fix is. But as XIO lives outside of the Nordstrom org and we have deadlines fast approaching, we felt think may be quicker to get merged. 

I'm happy to work with some members of the Platform team to evolve XRPC.

jkinkead: XIO is outside of Nordstrom, but it's run primarily by Nordstrom engineers. Getting a fix in should be fairly easy...

PR #206, opened 2018-09-18
Adds volume to pods

castillobgr:
Specifies an [`emptyDir`](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) volume in deployment.yml, needed because when Netty fails to load at least one `netty_tcnative*` lib variant, it generates one (to `/tmp` by default, but configurable at start-up through `-Dio.netty.native.workdir`). This can't be done in a read-only file system, such as one in pods under a restricted [Pod Security Policy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) with `ReadOnlyRootFilesystem: true` in k8s.

Mounting this volume as an `emptyDir` should solve this, unless the PSP forbids all use of volumes, or allows read-only volumes alone.

### DIFF
--- a/k8s-resources/deployment.yaml
+++ b/k8s-resources/deployment.yaml
@@ -18,6 +18,9 @@ spec:
       labels:
         app: xrpc
     spec:
+      volumes:
+      - name: netty-tmp-volume # Required for Netty to be able to export .so files under restrictive pod security policies
+        emptyDir: {}
       containers:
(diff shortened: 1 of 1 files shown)


castillobgr:
@mariusgrigoriu here's the PR :smile:

andyday:
👍 LGTM!  Merging this.

